Issue #4, opened 2025-06-10
CI Setup


igaray:
Set up the GH action CI jobs following the [miden-vm](https://github.com/0xMiden/miden-vm/tree/next/.github/workflows) and [miden-base](https://github.com/0xMiden/miden-base/tree/next/.github/workflows) repos. 

- Include the build, lint, and test flows.
- For the build job skip the no-std section.
- Skip the changelog flow for now.
- Set up an empty mdbook in the `docs` folder and use the books flow to publish the rendered version of the mdbook. 
- Set up another flow  called `publish-manifest`, triggered on push to main of modifications to the manifest files to publish the manifest to github pages.
- 
Once The CI config is merged we'll set up the branch protections.

PR #5, opened 2025-06-10
feat: add ci

lima-limon-inc:
Closes: #4 

This PR adds github actions for building, testing and linting based on the Actions present in `miden-vm` and `miden-base`. In addition, it also adds a job for publishing the midenup manifest to a Github Page; but only if the file itself is updated.


.github/workflows/publish-manifest.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  upload-manifest:
+    name: upload manifest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: updload manifest.json
```
igamigo, line 18: typo: 
```suggestion
      - name: Upload manifest.json
```

Makefile
```diff
@@ -0,0 +1,36 @@
...
+clippy: ## Runs Clippy with configs
+	cargo clippy -- -D warnings
+
+.PHONY: fix
+fix: ## Runs Fix with configs
+	cargo +nightly fix --allow-staged --allow-dirty --all-targets
+
+.PHONY: format
+format: ## Runs Format using nightly toolchain
+	cargo +nightly fmt --all
```
igamigo, line 13: Should we add a reference [`rustfmt.toml`](https://github.com/0xMiden/compiler/blob/next/rustfmt.toml)? (here's [`miden-base`'s](https://github.com/0xMiden/miden-base/blob/next/rustfmt.toml) for comparison)

igamigo (commented):
Leaving some

.github/workflows/publish-manifest.yml
lima-limon-inc: Great catch!  Thanks. Changed in this commit: https://github.com/0xMiden/midenup/pull/5/commits/c7678dacd5c3c4027c56cb9cb20b83b5e8af41ce

Makefile
lima-limon-inc: Hi Ignacio! Thanks I added `miden-base`'s `rustfmt.toml` file in this commit: https://github.com/0xMiden/midenup/pull/5/commits/a2c7ad730317c0ec96e4373e4590fed103eabc88

Sidenote: 
With that new `rustfmt.toml` file, the `lint` job was failing so a call to`make format` call was needed to adjust it to the new format. This was done here: https://github.com/0xMiden/midenup/pull/5/commits/414e9dda375aaefc0a358771be676e8c9e81c2b2

.github/workflows/lint.yml
bitwalker, line 1: I'm not sure we want to split up `build` and `lint` workflows like this, all of their jobs can run as part of the same workflow, since they are triggered on the same criteria, use the same toolchains, etc. I'm also not 100% sure that the `lint` workflow would pick up the cache used by the `build` workflow when separated like this, but that's a secondary concern.

We can save ourselves a ton of CI time by making the `build` job predicated on the `lint` jobs too, for reference, see how the compiler manages its CI workflow. If we forget to run the formatter, or forgot to address a clippy lint, the workflow is cancelled, rather than spending a bunch of CI cycles running the build and tests when that is going to get re-run anyway once the lints are fixed.

.github/workflows/test.yml
bitwalker, line 1: As mentioned in the other file, the jobs from the `test` workflow can be merged with the jobs of the `build` and `lint` workflows (and predicated on the `lint` jobs passing), since they all have the exact same configuration. I'd merge all three workflow files into one simply called `ci.yml`.

Cargo.toml
```diff
@@ -18,3 +17,15 @@ chrono = { version = "0.4.41", default-features = false, features = [
     "now",
 ] }
 toml = { version = "0.8", features = ["preserve_order"] }
+
+[profile.optimized]
```
bitwalker, line 21: Let's just modify the `release` profile directly, it's unnecessary to have both `release` and `optimized` profiles IMO.

Cargo.toml
```diff
@@ -18,3 +17,15 @@ chrono = { version = "0.4.41", default-features = false, features = [
...
+inherits = "release"
+codegen-units = 1
+lto = true
+
+[profile.test-dev]
```
bitwalker, line 26: I'm not sure I understand the need for this - it made sense in `miden-vm`, where enabling optimizations for the test suite is critical for performance, but we definitely don't need that here.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+.PHONY: format-check
+format-check: ## Runs Format using nightly toolchain but only in check mode
+	cargo +nightly fmt --all --check
+
+.PHONY: lint
+lint: format fix clippy ## Runs all linting tasks at once (Clippy, fixing, formatting)
```
bitwalker, line 20: I'd be a bit surprised if running `make lint` made changes to the source tree (which is what the `fix` task will do if it finds something to fix), I'd suggest we omit that task here.

Makefile
```diff
@@ -0,0 +1,36 @@
+# -- linting --------------------------------------------------------------------------------------
+
+.PHONY: clippy
+clippy: ## Runs Clippy with configs
+	cargo clippy -- -D warnings
+
+.PHONY: fix
+fix: ## Runs Fix with configs
+	cargo +nightly fix --allow-staged --allow-dirty --all-targets
```
bitwalker, line 9: I don't think we should be running `fix` using the nightly toolchain - it could result in fixes being applied that only work with the nightly toolchain, not stable.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+format-check: ## Runs Format using nightly toolchain but only in check mode
+	cargo +nightly fmt --all --check
+
+.PHONY: lint
+lint: format fix clippy ## Runs all linting tasks at once (Clippy, fixing, formatting)
+
+# --- testing -------------------------------------------------------------------------------------
+
+.PHONY: test-build
+test-build: ## Build the test binary
```
bitwalker, line 25: IMO we don't need this task at all, but I'm fine with keeping it for consistency with other projects for those that use it in their workflow. In any case, we can remove the explicit `--cargo-profile` flag here (since the `test-dev` profile should be removed)

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	cargo nextest run --cargo-profile test-dev --no-run
+
+
+.PHONY: test
+test: ## Run all tests
+	cargo nextest run --profile default --cargo-profile test-dev
```
bitwalker, line 31: We don't need the explicit `--profile` or `--cargo-profile` flags here, since the default profile is used by, well, default; and the `test-dev` profile can be removed.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+# --- building ------------------------------------------------------------------------------------
+.PHONY: build
+build: ## Builds with default parameters
+	cargo build --release
```
bitwalker, line 36: I don't think we want to build in `release` mode for this task, since it is primarily useful for doing local debugging. Only when we're installing the binary, i.e. with a `make install` task, or with `cargo-install` (where `--release` is the default anyway), would we want to ensure the resulting binary is built with in release mode.

manifest/channel-manifest.toml
bitwalker, line 1: I'm curious on the rationale for this change, I see two main downsides:

1. TOML is designed to be hand-written, which makes it a poor target for generating/modifying the manifest programmatically
2. We're planning to serve this up via GitHub Pages, and any number of tools might be built to pull and analyze it. JSON feels like a better format for that purpose, particularly because it is a supported content-type, whereas TOML is not.

Personally, I also think that as the number of channels grows, the TOML format here is going to become quite difficult to parse visually, and so the readability benefits won't be as stark as they are when we just have a single channel and a few components.

Basically, I think it might be premature to optimize for hand-maintaining the manifest, when I'm hoping we can move towards making the manifest largely maintained automatically (e.g. using `workflow_dispatch` jobs in GitHub Actions, or something along those lines). Was there any particular reason why you made the switch other than readability?

bitwalker (changes requested):
Thanks for taking the time to do this!

I have a few changes I'd like to make. I think using `miden-vm` and `miden-base` as starting points is fine, but they also have a lot of unnecessary complexity that simply doesn't translate to `midenup` (for sure `miden-vm`, but IIRC `miden-base` as well), so I think we can simplify things when starting from them. 

FWIW, the `compiler` repo is probably a better starting point, but I'm probably biased 😅. It also benefits from using `cargo-make` rather than `make`, so our automation is a lot cleaner; but that's different than how the other repos handle things, so I don't mind following their conventions outside of the compiler.

Anyway, happy to discuss any of the comments that are more subjective in nature, I'm opinionated but not fixated 😛

lima-limon-inc:
> Thanks for taking the time to do this!

On the contrary, thank you for your thorough and kind comments. 

> I have a few changes I'd like to make. I think using `miden-vm` and `miden-base` as starting points is fine, but they also have a lot of unnecessary complexity that simply doesn't translate to `midenup` (for sure `miden-vm`, but IIRC `miden-base` as well), so I think we can simplify things when starting from them.

Noted, I'll follow up on your notes in the comments.

> FWIW, the `compiler` repo is probably a better starting point, but I'm probably biased 😅. 

I can't blame you, the people behind the miden compiler have great taste 😉.

> It also benefits from using `cargo-make` rather than `make`, so our automation is a lot cleaner; but that's different than how the other repos handle things, so I don't mind following their conventions outside of the compiler.

My main motivation behind using `make` was to follow the other repositories convention and also because it's a tool I'm more familiar with. 
However, do note that it can be changed to `cargo-make` if you find it a more suitable tool.

manifest/channel-manifest.toml
lima-limon-inc: Great points.

> Was there any particular reason why you made the switch other than readability?

The motivation was in part readability, but mainly to follow the rust foundation's [channel layout](https://forge.rust-lang.org/infra/channel-layout.html), which uses toml as a format. Additionally, one possible benefit of `toml` as a format is that it is largely append only. The rationale ended mostly there however.

I reverted this change in this commit: https://github.com/0xMiden/midenup/pull/5/commits/896b14fad97aac4eab77be6d94deec9852ae3c87

Cargo.toml
lima-limon-inc: Noted. Changed here: https://github.com/0xMiden/midenup/pull/5/commits/4f3a320a43086062a62bd2cc557d6d6463fa9366

Additionally, I also modified the the profile's settings to be more aligned with the compiler's, here: https://github.com/0xMiden/midenup/pull/5/commits/89747dbdea7fa51bfc070fa8d9e6f12a0af619f5. 
Should this change stay or should it be reverted?

Makefile
lima-limon-inc: Noted. I removed the flag in this commit: https://github.com/0xMiden/midenup/pull/5/commits/84ef58883f0bf15dea20e06cacd89c3b28b78382

Cargo.toml
lima-limon-inc: Removed here: https://github.com/0xMiden/midenup/pull/5/commits/0955d5b54ce93dd6f3ba67e467032d9c9bdd53e6

.github/workflows/test.yml
lima-limon-inc: Addressed here: https://github.com/0xMiden/midenup/pull/5#discussion_r2140960417

Makefile
lima-limon-inc: Changed here: https://github.com/0xMiden/midenup/pull/5/commits/67d9b4dc82a8bf32915a1b3bf9585571a02a22af

Makefile
lima-limon-inc: Noted, changed here: https://github.com/0xMiden/midenup/pull/5/commits/cad409a288317aa7f1f2d3cad15c06ab23a4d03e

.github/workflows/lint.yml
bitwalker: We can have the `build` and `test` jobs run in parallel, since the builds use different profiles, and if the regular `build` job would fail, then so would the `test` job, but in the happy path case, both would run concurrently and speed up CI runs when the author has already run all the checks locally.

Cargo.toml
bitwalker: I think we should probably remove the `lto = "thin"` here and in the compiler, since it makes no sense to use ThinLTO when `codegen-units = 1` (since the purpose of ThinLTO is to allow for LTO to work across codegen units).

Otherwise, looks good!

manifest/channel-manifest.toml
bitwalker: Ah, I see, yeah `rustup` seems to use a channel manifest per-channel, so I would imagine that counteracts one downside I was anticipating. Good point about TOML being append-only as well.

We can continue to re-evaluate our choice here in the near term, until we're pretty confident in the set of constraints the format needs to handle. I think JSON is still the best starting point, but I wouldn't be opposed to switching to TOML in the future if it turns out the benefits of JSON I outlined aren't really needed.

bitwalker:
> My main motivation behind using make was to follow the other repositories convention and also because it's a tool I'm more familiar with.
However, do note that it can be changed to cargo-make if you find it a more suitable tool.

Definitely not that important in the grand scheme of things. We use `cargo-make` in the compiler because it has much more sophisticated support for Rust projects in particular, with baked in support for automatically installing third-party Cargo extensions when they are referenced, knowledge about Rust workspaces vs crates, the ability to execute Rust scripts as tasks, that sort of thing. Beyond that, the ability to pass arguments to tasks (and make use of them in task definitions), and finer-grained control over the task dependency graph, are really nice features that I wish `make` itself had.

In any case, for simple projects, `cargo-make` is overkill, for the compiler it's great (and I'd argue it'd be really nice for the VM and some of the other complex components, but "if it ain't broke, don't fix it").

bitwalker (approved):

Issue #9, opened 2025-06-12
`midenup install` should store the manifest it used in `$MIDENUP_HOME`


bitwalker:
This is probably a blocker for #8.

In order to determine what is needed for `update`, we need to compare the versions of the components we installed against the latest versions in the published manifest, for a few reasons:

1. We can avoid doing redundant work if no changes have occurred to the set of installed components
2. We can handle the case where a toolchain was yanked for some reason (i.e. we identified that there was some incompatibility, or other significant issue not fixed by simply releasing a new version of a single component)
3. We can determine what the meaning of `stable` or `nightly` was at the time of installation, and be able to determine whether or not we need to install a new toolchain, or just update individual components.

The manifest isn't _exactly_ the same as the published one, instead it needs to represent the set of installed toolchains and components, specifically the versions and features that were used at the time of installation. When `update` is run and changes are applied, we should then update the local manifest as appropriate (i.e. if you run `midenup update stable`, we'd update the information for the `stable` toolchain that is installed, but leave the rest of the manifest untouched).

We should write the local manifest to `$MIDENUP_HOME/manifest.json`.

PR #11, opened 2025-06-12
refactor(channel): make stable an alias for latest version

lima-limon-inc:
Currently `stable` is treated as a channel on its own, after this PR, stable will become an alias for the latest version. To achieve this, the `ChannelType` enum will be split:

- `UserChannel` enum: Which represents the "user facing channels", i.e. "stable", "nightly" or a specific version. 
- `ChannelAlias` enum: Which represents an optional tag which will be used in the local manifest. These will be used to mark the local state of the manifest, and will be used to update the toolchain as described here: https://github.com/0xMiden/midenup/issues/9.

This changes: 
- The `channel_manifest.json`: Now there is no "stable" channel, only versioned channels.
- After installing `stable`, the `stable` directory is created as a symbolic link pointing to the corresponding version.


src/channel.rs
```diff
@@ -50,6 +50,87 @@ impl Component {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged, rename_all = "snake_case")]
+pub enum CanonicalChannel {
+    /// This channel represents the latest nightly versions of all components
+    Nightly,
+    /// This channel represents the latest stable versions of all components compatible with
+    /// the specified toolchain version string.
+    Version {
+        version: semver::Version,
+        #[serde(skip_serializing)]
```
bitwalker, line 63: I would suggest changing this to `is_nightly`. Then, when looking up the "stable" alias, we just order all of the versions sans those where `is_nightly` is true and take the greatest version. The same could be done to resolve the latest nightly as well, just for the inverse value of that flag.

src/manifest.rs
```diff
@@ -58,30 +58,74 @@ impl Manifest {
...
+                ) => x.cmp_precedence(y),
+            });
+
+        if let Some(stable) = stable_channel {
+            stable.name = match &stable.name {
+                CanonicalChannel::Nightly => CanonicalChannel::Nightly,
+                CanonicalChannel::Version { version: x, .. } => {
+                    CanonicalChannel::Version { version: x.clone(), is_stable: true }
+                },
+            }
```
lima-limon-inc, line 88: This parses the received manifest and marks the largest parsed version as `stable`.

The main goal behind this is for Midenup to only have two distinct types of Channels `versions` and `nightly`. Then, the different  "user facing channels" i.e. `"stable"`, `version` and `"nightly"` need only be mapped to those internal channels.

I don't think this can be done while parsing, since you need the entirety of the versions to determine which one of them is stable.

src/channel.rs
lima-limon-inc: > I would suggest changing this to `is_nightly`. Then, when looking up the "stable" alias, we just order all of the versions sans those where `is_nightly` is true and take the greatest version.

My original intention with `CanonnicalChannel::Version` was for it to  unify both `stable` and `version`.   Thus, `stable`, under the hood, is just another `CanonnicalChannel::Version` that gets flagged as `is_stable` during initialization (done [here](https://github.com/0xMiden/midenup/pull/11/files#diff-a44cbf177b0e747788be2cc21913a967ce8a76f371b98a6c17cd908c400bf1a9R68-R88)).


With this in mind, obtaining the stable channel is [done like so](https://github.com/0xMiden/midenup/pull/11/files#diff-a44cbf177b0e747788be2cc21913a967ce8a76f371b98a6c17cd908c400bf1a9R105-R112). What do you think about this approach?

bitwalker: Thinking about this a bit further, I'm wondering how you feel about the following:

```rust
/// Internal representation of a channel
pub struct Channel {
    version: semver::Version,
    alias: Option<ChannelAlias>,
}

impl Channel {
    /// Is this channel a stable release? Does not imply that it has the `stable` alias
    pub fn is_stable(&self) -> bool {
        self.alias.as_ref().is_none_or(|alias| matches!(alias, ChannelAlias::Stable))
    }

    /// Is this channel the current version carrying the `stable` alias
    pub fn is_latest_stable(&self) -> bool {
        self.alias.as_ref().is_some_and(|alias| matches!(alias, ChannelAlias::Stable))
    }

    pub fn is_nightly(&self) -> bool {
        self.alias.as_ref().is_some_and(|alias| matches!(alias, ChannelAlias::Nightly(_)))
    }

    pub fn is_latest_nightly(&self) -> bool {
        self.alias.as_ref().is_some_and(|alias| matches!(alias, ChannelAlias::Nightly(None)))
    }
}

pub enum ChannelAlias {
    /// Represents `stable`
    Stable,
    /// Represents either `nightly` or `nightly-$SUFFIX`
    Nightly(Option<Cow<'static, str>>),
    /// An ad-hoc named alias for a channel
    Tag(Cow<'static, str>),
}

/// User-facing channel reference
#[serde(untagged, rename_all = "snake_case")]
pub enum UserChannel {
    // This variant is tried first, then stable, then nightly, then fallback
    Version(semver::Version),
    Stable,
    Nightly,
    Other(Cow<'static, str>),
}

pub struct Manifest {
    ...,
    channels: Vec<Channel>,
}

impl Manifest {
    pub fn get_latest_stable(&self) -> Option<&Channel> {
        self.channels.iter().find(|c| c.is_latest_stable())
            .or_else(|| {
                self.channels.iter().filter(|c| c.is_stable()).max_by(|x, y| x.version.cmp_precedence(&y.version))
            })
    }

    pub fn get_channel(&self, channel: &UserChannel) -> Option<&Channel> {
        match channel {
            UserChannel::Version(v) => self.channels.iter().find(|c| &c.version == v),
            UserChannel::Stable => self.get_latest_stable(),
            UserChannel::Nightly => self.get_latest_nightly(),
            UserChannel::Other(tag) => match tag.split_prefix("nightly-") {
                Some(suffix) => self.get_named_nightly(suffix),
                None => self.channels.iter().find(|c| c.alias.is_some_and(|alias| matches!(alias, ChannelAlias::Tag(t) if t == tag)),
            },
        }
}
```

With this setup, the manifest always contains the "canonical" representation of a channel; and we can trivially resolve a user-provided channel name/alias to a canonical channel via the manifest.

I think this ultimately ends up being a lot cleaner. What's your take?

src/channel.rs
```diff
@@ -11,7 +11,7 @@ use crate::version::Authority;
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Channel {
     /// The name of the channel
-    pub name: ChannelType,
+    pub name: CanonicalChannel,
```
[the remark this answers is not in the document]
bitwalker: See my other comment where I rethink the design a bit, it addresses the thing I was after here in the process

src/channel.rs
lima-limon-inc: > I understand the rationale, but I think aliasing is always fundamentally optional. The notion of `SpecificRelease` is already represented by the version field itself.
> 
> Additionally, most channels will just be versions, not aliased, so requiring the field in the manifest would be unnecessary noise IMO.

Noted, good point.

> I expect that it will be used for local development toolchains (i.e. when working on Miden itself). These would be given names/tags for convenience. It's something I want to experiment with soon, since it's a use case I personally have, but am not yet sure what the workflow should look like exactly, but named toolchains like this is something I expect to be part of it.

I see, thanks for the context. 

I'll get to work on this new approach.

bitwalker: > Quick question, just so we are on the same page, this approach implies adding the alias in the channel_manifest.json file itself, right?

Yes, exactly, at least in the one that we maintain locally after `midenup install`. The upstream toolchain manifest doesn't necessarily need to provide the alias explicitly, it would just be resolved dynamically in that case. That allows us to programmatically update the upstream `channel-manifest.json` without having to compute toolchain aliases.

src/channel.rs
```diff
@@ -21,6 +23,66 @@ impl Channel {
         let name = name.as_ref();
         self.components.iter().find(|c| c.name == name)
     }
+    /// Is this channel a stable release? Does not imply that it has the `stable` alias
```
lima-limon-inc, line 26: Note: `is_latest_stable` got removed because of [this comment](https://github.com/0xMiden/midenup/pull/11#discussion_r2154899408)

bitwalker (approved):

PR #15, opened 2025-06-17
feat(CI): create CI job to test intalling stable toolchain

### LINKED ISSUE #14: Installing stable

lima-limon-inc:
Closes #14 

Add a small CI job to check the `stable` toolchain can be installed via `midenup`.

### DIFF
--- a/src/commands/install.rs
+++ b/src/commands/install.rs
@@ -98,7 +98,7 @@ fn main() {
     // Install {{ component.name }}
     let mut child = Command::new("cargo")
         .arg(
-          "{{ component.required_toolchain }}",
+          "{{ component.required_toolchain_flag }}",
         )
         .arg("install")
         .args([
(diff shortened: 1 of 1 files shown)


lima-limon-inc:
> Looks good! Let's move this to a new workflow though, and predicate the job on a label being added called `check:install` (I'll create it now). That way we can avoid saturating CI unless a PR rises to the level of verifying things end-to-end.

I changed it here: https://github.com/0xMiden/midenup/pull/15/commits/4c92e561a4eacc80f06e989f7d5b1f428e7ced8b

Btw, in this commit https://github.com/0xMiden/midenup/pull/15/commits/bb89c461144f98710fb5a2a6674a936d621eb9dd I configured the release profile to be run for this workflow. And here https://github.com/0xMiden/midenup/pull/15/commits/41d98c34428f0ee9b406d5aa2e112dba724e93ed I made it so the installing stable workflow is run on main.

What do you think of these? Should I revert them?

bitwalker (approved):

Issue #8, opened 2025-06-12
Implement `midenup update`


lima-limon-inc:
Currently `midenup update`'s implementation is missing. 

Side-note: I am assuming this is blocked by PR #7 which changes how installation is handled.

PR #17, opened 2025-06-26
feat: implement midenup update

lima-limon-inc:
Closes: #8


.github/workflows/ci.yml
```diff
@@ -58,15 +58,39 @@ jobs:
   test:
     name: test
     runs-on: ubuntu-latest
+    env:
```
bitwalker, line 61: I think I'd be more inclined to just cache things like we do in other projects (e.g. the compiler), rather than bring `sccache` into the mix, since I don't think we're likely to see significant benefits doing so versus the simpler approach.

Either way, it's key that we're able to bust the cache manually, as we inevitably will need to do so at some point.

.github/workflows/ci.yml
```diff
@@ -58,15 +58,39 @@ jobs:
...
+      - uses: mozilla-actions/sccache-action@v0.0.9
+
       - uses: actions/checkout@main
+
+      - name: Restore Sccache
+        id: cache-sccache-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.cache/sccache/
+          key: ${{ github.event.number }}-sccache
```
bitwalker, line 74: I'm not actually sure what the `github.event.number` value is derived from. Does it change frequently, or very infrequently? If it changes on every push, then we're never going to hit a hot cache; but if it never changes, then we don't have a reliable way of flushing stale stuff from the cache.

See the other Miden projects (for sure the compiler, and I think the VM also uses something on par with what the compiler does) for how we generally handle caching for Rust builds to see what we use for cache keys/cache busting.

src/channel.rs
```diff
@@ -40,6 +40,27 @@ impl Channel {
             .as_ref()
             .is_some_and(|alias| matches!(alias, ChannelAlias::Nightly(None)))
     }
+
+    pub fn components_to_update(&self, newer: &Self) -> Vec<Component> {
+        // NOTE: This wouldn't work if they have different amount of elements
```
bitwalker, line 45: You can construct a `BTreeSet`/`HashSet`, and:

1. `prev.difference(new)` gives you the set of removed components
2. `new.difference(prev)` gives you the set of added components
3. `prev.intersection(new)` gives you the set of components present in both prev and new sets, and thus either the set of components to update, or at least the set of components to check if an update is required

src/commands/install.rs
```diff
@@ -1,32 +1,43 @@
...
+) -> anyhow::Result<()> {
     config.ensure_midenup_home_exists()?;
 
     let installed_toolchains_dir = config.midenup_home.join("toolchains");
     let toolchain_dir = installed_toolchains_dir.join(format!("{}", &channel.name));
 
-    if toolchain_dir.exists() {
+    // NOTE: The installation indicator is only created after successfull
+    // toolchain installation.
+    let installation_indicator = toolchain_dir.join("installation-successfull");
```
bitwalker, line 26:
```suggestion
    let installation_indicator = toolchain_dir.join("installation-successful");
```

src/commands/update.rs
```diff
@@ -1,6 +1,146 @@
...
+
+                remove_exe
+                    .wait()
+                    .context("failed to uninstall component '{{ component.name }}'")?;
+            },
+            _ => todo!(),
+        }
+    }
+
+    // NOTE: After deleting the files we need to remove the "all is installed
```
bitwalker, line 136: I would remove this file first actually - that way if something goes wrong, a subsequent `install`/`update` will reinstall everything. If we remove it last, we might remove various components, hit an error part way through, and then bail before removing the install marker.

bitwalker (commented):
Looks great! Nice work!

A few comments/nits, but aside from those things I think this is ready to be merged!

src/commands/install.rs
```diff
@@ -127,42 +134,71 @@ fn main() {
...
+        // Await results
+        child.wait().expect("failed to install component '{{ component.name }}'");
+    }
 
-    // Await results
-    child.wait().expect("failed to install component '{{ component.name }}'");
+    {% endfor %}
 
+    // This file indicates that installation finished successfully
+    let checkpoint_path = miden_sysroot_dir.join("installation-successfull");
```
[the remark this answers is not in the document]
lima-limon-inc: My bad! Fixed here: https://github.com/0xMiden/midenup/pull/17/commits/9187db2bc4767f2db1342c998909af5e562d724d

.config/nextest.toml
```diff
@@ -0,0 +1,7 @@
+[test-groups]
+serial-integration = { max-threads = 1 }
+
+[[profile.default.overrides]]
+filter = 'test(integration)'
+test-group = 'serial-integration'
+
```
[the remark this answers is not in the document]
lima-limon-inc: Sidenote: Particularly for these midenup tests, they could be potentially run in parallel if each test is run on a separate github runner. With a setup like that, we wouldn't bump into `rustup` limitation with concurrent execution.

This should speed up CI execution; although I can't think of an ergonomic way to configure this without needing to name tests explicitly. The first thing that comes to mind are [strategy matrices](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow).
I don't think it's a problem for now, but maybe something to keep in mind.

.github/workflows/ci.yml
bitwalker: Ah, I see. That _feels_ like too ephemeral a cache to be worthwhile to me, but without concrete numbers, I honestly can't say. I'm fine with testing it out for awhile though - it'd definitely be nice to gather some stats on it, but we'll probably have to run it for a bit to do that.

src/channel.rs
lima-limon-inc: Hi! So I implemented this functionality in this commit: https://github.com/0xMiden/midenup/pull/17/commits/f7cafe83f267aa0b8a4dbd4c13bd379abaf78769

What do you think about the implementation? I was hesitant about there being both a "are_equal" function and a "PartialEq" implementation. Maybe there's a simpler way that I missed; apologies in advance if that's the case.

.github/workflows/ci.yml
lima-limon-inc: > Ah, I see. That _feels_ like too ephemeral a cache to be worthwhile to me, but without concrete numbers, I honestly can't say. I'm fine with testing it out for awhile though - it'd definitely be nice to gather some stats on it, but we'll probably have to run it for a bit to do that.

Noted! Maybe it can be tackled later on if CI times get out of hand. I could create an issue to track it if you feel like it merits it.

bitwalker (approved):

PR #27, opened 2025-07-02
feat: add the client to the manifest + `Authority::Git` implementation

### LINKED ISSUE #26: feat: add miden client to the manifest
Ensure the channel manifest includes the miden client version and that the build script installs the correct version of miden client.

### LINKED ISSUE #35: feat: make sure that `Authority::Git` is working and add tests
Currently, the `Authority` enum does contain a `Git` variant; however it hasn't been tested much. 

- [ ] Ensure installation works
- [ ] Ensure updating works
- [ ] Add tests
- [ ]  Enhance the variant by adding a `Hash` field to force the use of a specific commit hash.

lima-limon-inc:
Closes #26 
Closes #35 

This PR adds the Miden client to the manifest and also adds the implementation for components behind `Authority::Git`

Notes on this PR:
- The manifest now holds the `miden-client`, but it is tracked via a specific commit. This is done in order to guarantee a component that is not susceptible to changes (since it is not a desired behavior).   
- `Authority::Git` supports three different targets: tags, revisions and branches
- When an update is triggered, components behind branches will check if there have been any new commits since (this is done via the `git ls-remote` command). If so, an update is issued.

### DIFF
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -24,9 +24,8 @@
         },
         {
           "name": "miden-client",
-          "package": "miden-cli",
-          "version": "0.9.3",
-          "installed_file": "miden"
+          "repository_url": "https://github.com/0xMiden/miden-client.git",
+          "installed_file": "miden-client"
         },
         {
           "name": "midenc",
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -25,6 +25,7 @@
         {
           "name": "miden-client",
           "repository_url": "https://github.com/0xMiden/miden-client.git",
+          "crate_name": "miden-cli",
           "installed_file": "miden-client"
         },
         {
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -26,8 +26,7 @@
           "name": "miden-client",
           "repository_url": "https://github.com/0xMiden/miden-client.git",
           "revision": "83df2aa115b2617e211d1d929a1189ffabbd137b",
-          "crate_name": "miden-client-cli",
-          "installed_file": "miden-client-cli"
+          "crate_name": "miden-client-cli"
         },
         {
           "name": "midenc",
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -25,7 +25,7 @@
         {
           "name": "miden-client",
           "repository_url": "https://github.com/0xMiden/miden-client.git",
-          "branch": "fabrizioorsi/midenup-exe-patch",
+          "revision": "83df2aa115b2617e211d1d929a1189ffabbd137b",
           "crate_name": "miden-client-cli",
           "installed_file": "miden-client-cli"
         },
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -25,8 +25,9 @@
         {
           "name": "miden-client",
           "repository_url": "https://github.com/0xMiden/miden-client.git",
-          "crate_name": "miden-cli",
-          "installed_file": "miden-client"
+          "branch": "fabrizioorsi/midenup-exe-patch",
+          "crate_name": "miden-client-cli",
+          "installed_file": "miden-client-cli"
         },
         {
           "name": "midenc",
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -23,6 +23,12 @@
           "installed_file": "miden"
         },
         {
+          "name": "miden-client",
+          "package": "miden-cli",
+          "version": "0.9.3",
+          "installed_file": "miden"
+        },
+        {
           "name": "midenc",
           "version": "0.1.0",
           "rustup_channel": "nightly-2025-03-20"
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -46,6 +46,8 @@ impl Channel {
             .is_some_and(|alias| matches!(alias, ChannelAlias::Nightly(None)))
     }
 
+    /// This functions compares the Channel &self, with a newer channel [newer]
+    /// and returns the vector of elements that need to be updated.
     pub fn components_to_update(&self, newer: &Self) -> Vec<Component> {
         let new: HashSet<&Component> = HashSet::from_iter(newer.components.iter());
         let current: HashSet<&Component> = HashSet::from_iter(self.components.iter());
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -344,10 +344,16 @@ impl core::str::FromStr for UserChannel {
 mod tests {
     use crate::{
         channel::{Channel, Component},
-        version::Authority,
+        version::{Authority, GitTarget},
     };
 
     #[test]
+    /// This tests checks that the [Channel::components_to_update] functions behaves as intended.
+    /// Here the following updates need to be performed:
+    /// - vm requires update 0.12.0 -> 0.15.0
+    /// - std requires downgrade from 0.15.0 -> 0.12.0
+    /// - a so called "removed-component" needs to be deleted
+    /// - a so called "new-component" needs to be added
     fn check_components_to_update() {
         let old_components = [
             Component {
@@ -374,8 +380,6 @@ mod tests {
             },
             Component {
                 name: std::borrow::Cow::Borrowed("removed-component"),
-                // This should be present in the update vector
-                // Component that got removed
                 version: Authority::Cargo {
                     package: Some(String::from("deleted-repo")),
                     version: semver::Version::new(0, 82, 77),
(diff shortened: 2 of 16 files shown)


lima-limon-inc:
Sidenote: I think this is blocked until the names of the executables are made to be unique.

I originally attempted to work my way around this limitation; but there's a problem with `cargo` itself: If two (or more) crates generate executables with the same name, the executables are *overwritten*, *without warning*.

This can be seen in the *crates.toml* file generated by `cargo` after installing a package. The contents of said file **before** installing the client was the following:

```
[v1]
"cargo-miden 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)" = ["cargo-miden"]
"miden-vm 0.15.0 (registry+https://github.com/rust-lang/crates.io-index)" = ["miden"]
```

**After** the client got installed, it read:

```
[v1]
"cargo-miden 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)" = ["cargo-miden"]
"miden-cli 0.9.3 (registry+https://github.com/rust-lang/crates.io-index)" = ["miden"]
```

This happened, even though the executable is manually renamed via `std::fs::rename`.

manifest/channel-manifest.json
```diff
@@ -22,6 +22,13 @@
           "features": ["executable", "concurrent"],
           "installed_file": "miden"
         },
+        {
+          "name": "miden-client",
+          "repository_url": "https://github.com/0xMiden/miden-client.git",
+          "branch": "fabrizioorsi/midenup-exe-patch",
+          "crate_name": "miden-client-cli",
+          "installed_file": "miden-client-cli"
```
lima-limon-inc, line 30: [fabrizioorsi/midenup-exe-patch](https://github.com/lambdaclass/miden-client/tree/fabrizioorsi/midenup-exe-patch) is a branch in the miden-client repository. It stems from main, only adding the commit the renamed the client's executable to `miden-client-cli`. 

The branch is only temporary however, I believe that as soon as that change gets merged to main, it can be safely deleted.

src/main.rs
```diff
@@ -165,9 +169,14 @@ fn main() -> anyhow::Result<()> {
...
-                "new" => ("cargo", vec!["miden", "new"]),
-                other => (other, vec![]),
+                "build" => {
+                    (String::from("cargo"), vec![String::from("miden"), String::from("build")])
+                },
+                "new" => (String::from("cargo"), vec![String::from("miden"), String::from("new")]),
+                other => {
+                    let command = format!("miden-{other}");
+                    (command, vec![])
+                },
```
lima-limon-inc, line 179: These changes are already in `main`, they were merged in https://github.com/0xMiden/midenup/pull/34.

lima-limon-inc:
Sidenote: Both `Integration` tests fail because they are running against the published Manifest. 

Same issue that happened here: https://github.com/0xMiden/midenup/pull/11#issuecomment-2971697667

I'll create an issue for this

manifest/channel-manifest.json
```diff
@@ -22,6 +22,12 @@
           "features": ["executable", "concurrent"],
           "installed_file": "miden"
         },
+        {
+          "name": "miden-client",
+          "package": "miden-cli",
```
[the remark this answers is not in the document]
igamigo: The rename has already been applied on the `next` branch BTW, so the coming release will publish the crate as `miden-client-cli`

manifest/channel-manifest.json
```diff
@@ -22,6 +22,12 @@
           "features": ["executable", "concurrent"],
           "installed_file": "miden"
         },
+        {
+          "name": "miden-client",
+          "repository_url": "https://github.com/0xMiden/miden-client.git",
+          "revision": "83df2aa115b2617e211d1d929a1189ffabbd137b",
```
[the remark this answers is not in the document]
bitwalker: I think the use case for referencing a git branch is sufficiently niche enough that it is a non-issue in practice that the branch can evolve in potentially breaking ways. Particularly since it requires an explicit `midenup update` anyway, I think it's likely this is the behavior one would expect.

src/channel.rs
```diff
@@ -46,26 +49,29 @@ impl Channel {
             .is_some_and(|alias| matches!(alias, ChannelAlias::Nightly(None)))
     }
 
+    /// This functions compares the Channel &self, with a newer channel [newer]
+    /// and returns the list of [Components] that need to be updated.
     pub fn components_to_update(&self, newer: &Self) -> Vec<Component> {
-        let new: HashSet<&Component> = HashSet::from_iter(newer.components.iter());
+        let new_channel: HashSet<&Component> = HashSet::from_iter(newer.components.iter());
         let current: HashSet<&Component> = HashSet::from_iter(self.components.iter());
```
bitwalker, line 56:
```suggestion
        let new_channel = HashSet::from_iter(newer.components.iter());
        let current = HashSet::from_iter(self.components.iter());
```

The types here can be inferred, so I would prefer to elide them. As a general rule, explicit types are worth adding when it isn't clear what the type _should_ be inferred as, or when the compiler can't infer them - the rest of time there's little reason to annotate them.

src/commands/install.rs
```diff
@@ -75,16 +75,50 @@ pub fn install(
     }
 
     // Update local manifest
+    // -------------------------------------------------------------------------
```
bitwalker, line 78: *Nit:* I'd prefer to not have line comments that add sections like this. If the code isn't clear without them, then we probably should refactor it; and if it is clear enough without them, then they aren't needed.

src/commands/install.rs
```diff
@@ -75,16 +75,50 @@ pub fn install(
...
+            modifiable.alias = Some(ChannelAlias::Stable);
+            modifiable
+        } else {
+            channel.clone()
+        };
+
+        // If a component was installed with --branch, then write down the
+        // current commit. This is used on updates to check if any new commits
+        // were pushed since installation.
+        // NOTE: To check the latest commit we're using git cli instead. Would
```
bitwalker, line 93: I think it is fine to expect `git` to be on the user's PATH, and just shell out to it. It'll be far simpler, and we don't need to effectively ship a git binary inside `midenup`.

bitwalker (approved):
Looks great! A couple of nits, but I'll mark this approved. Let me know when you're ready for it to be merged 🙂

src/channel.rs
lima-limon-inc: >  or when the compiler can't infer them 

I think this is case in this particular snippet. I committed the change to showcase the error, [here](https://github.com/0xMiden/midenup/actions/runs/16194237056/job/45716741067#step:6:133). (Sidenote: I'll remove this commit before merging).

I think this is the same problem as the one described in this [StackOverflow question](https://stackoverflow.com/q/62949404). This seems to be a problem with `from_iter`, here's a little recreation in the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3aa55152947fe266af1dc647107b2a4a

Interestingly enough, only one explicit type hint is needed to satisfy the compiler. The following does compile: 
```
    pub fn components_to_update(&self, newer: &Self) -> Vec<Component> {
        let new_channel: HashSet<&Component> = HashSet::from_iter(newer.components.iter());
        let current = HashSet::from_iter(self.components.iter());
```
Although, I decided against it originally because it seemed a bit inconsistent. 


With that in mind, I pushed this commit: https://github.com/0xMiden/midenup/pull/27/commits/47af5594ac0379107e337c4b9106a42e3f0b1a88 which removed the inner type of the `HashSet`.

Let me know which version you prefer!

bitwalker: Ah interesting! I suspect it's because of how we use `new_channel` vs `current`; if you try to interact with an object with an inferred type in a way that also requires inference, the compiler complains and wants you to specify a type. I think in this case, `current` avoids the need for the type annotation because we are passing it as an argument to a `HashSet` method that requires a specific concrete type, but only if the type of `new_channel` is known, so it complains about one but not the other.

I think it's fine to minimize the type annotations, even if it looks a bit weird to annotate one but not the other, it's essentially just to satisfy the compiler in this case.

lima-limon-inc: > Ah interesting! I suspect it's because of how we use `new_channel` vs `current`; if you try to interact with an object with an inferred type in a way that also requires inference, the compiler complains and wants you to specify a type. I think in this case, `current` avoids the need for the type annotation because we are passing it as an argument to a `HashSet` method that requires a specific concrete type, but only if the type of `new_channel` is known, so it complains about one but not the other.
> 
> I think it's fine to minimize the type annotations, even if it looks a bit weird to annotate one but not the other, it's essentially just to satisfy the compiler in this case.

Noted! Changed it here: https://github.com/0xMiden/midenup/commit/75107e41c834ec86428a3900e0c51452b83a7646. With this now in place, I think this is ready to get merged 😊

PR #34, opened 2025-07-03
fix: add `miden-` prefix to the subcommand passed to `miden`

lima-limon-inc:
This PR adds the `miden-` prefix to the subcommand passed to `miden` via the CLI. 

In addition, it also removes the hard-coded `file:://manifest/channel-manifest.json` manifest in favor of the save treatment present in `midenup`.


bitwalker (approved):

bitwalker (changes requested):
Sorry, meant to request that we retain setting the `PATH` directory as before

Issue #40, opened 2025-07-08
bug: midenup doesn't register if the install script fails


lima-limon-inc:
Currently, most invocations of `Command::spawn` call `child.wait().context(<>)` directly. 
The problem is, `child.wait` doesn't error out when the `Command` errors, but rather when it fails to *wait*.

Here's a recreation:
```
$~ tree .
.
├── Cargo.lock
├── Cargo.toml
└── src
    ├── main.rs
    └── script.rs

$~ cat src/main.rs
use std::collections::HashMap;
use std::process::Command;

fn main() {
    let mut child = Command::new("cargo")
        .args(["+nightly", "-Zscript"])
        .arg("src/script.rs")
        .stderr(std::process::Stdio::inherit())
        .stdout(std::process::Stdio::inherit())
        .spawn()
        .unwrap();

    let b = child.wait().unwrap();
}
$~ cat src/script.rs
#!/usr/bin/env cargo


use std::io::Write;
use std::process::Command;

fn main() {
    let mut a = Command::new("nope")
        .arg("-la")
        .stderr(std::process::Stdio::inherit())
        .stdout(std::process::Stdio::inherit())
        .spawn()
        .expect("failed to spawn");

    a.wait().unwrap();
}

```

If we run `cargo run`, the script will fail. It will even display the following error:
```
thread 'main' panicked at script.rs:13:10:
failed to spawn: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

However, *the exit code is 0*,
```
$~ echo $?
0
```

If I'm not mistaken, In order to check if the `Command` failed, we need to check this via the `success` method: https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.success and catch the error via that. Something along these lines:
```
use std::collections::HashMap;
use std::process::Command;

fn main() {
    let mut child = Command::new("cargo")
        .args(["+nightly", "-Zscript"])
        .arg("src/script.rs")
        .stderr(std::process::Stdio::inherit())
        .stdout(std::process::Stdio::inherit())
        .spawn()
        .unwrap();

    let b = child.wait();
    match b {
        Ok(a) => {
            if a.success() {
                println!("All good");
            } else {
                println!("Failed");
            }
        }
        Err(a) => todo!(),
    }
}
```

Issue #39, opened 2025-07-08
feat: make missing components to `Authority::Path`


lima-limon-inc:
This will enable midenup to handle components that are present in the filesystem, managed via `cargo install --path`.

All the functionality is basically in place, the main missing bit is how to handle updates. 
My initial thought is to simply skip components that are taken from `--path`, but would love to hear any thoughts!

PR #41, opened 2025-07-08
feat: catch install script errors + implement `Authority::Path` update

lima-limon-inc:
Closes: #40 
Closes: #39 

This PR expands the handling of error when `std::proccess::Command` is used. Some uses were missing a `.success()` check, which meant that if the install script failed, midenup failed to catch that error and simply continued. 
A `.success()` call was added in all of the missing places and a small test was added to check that midenup registers when a install fails.

### DIFF
--- a/src/commands/install.rs
+++ b/src/commands/install.rs
@@ -227,25 +227,8 @@ fn main() {
 
 
         if !status.success() {
-            let error = child.stderr.take();
-
-            let error_msg = if let Some(mut error) = error {
-                let mut stderr_msg = String::new();
-                let read_err_msg = error.read_to_string(&mut stderr_msg);
-
-                if read_err_msg.is_err() {
-                    String::new()
-                } else {
-                    format!("The following error was raised: {stderr_msg}")
-                }
-            } else {
-                String::new()
-            };
-
             panic!(
-                "midenup failed to uninstall '{{ component.name }}' with status {}. {}",
-                status.code().unwrap_or(1),
-                error_msg
+                "midenup failed to uninstall '{{ component.name }}'"
             );
         }
     }
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -1,5 +1,3 @@
-use std::io::Read;
-
 use anyhow::{Context, bail};
 
 use crate::{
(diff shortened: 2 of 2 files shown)


lima-limon-inc:
IINM, the `Integration` tests should get fixed after #27 gets merged

src/commands/install.rs
```diff
@@ -213,7 +223,31 @@ fn main() {
             .expect("failed to install component '{{ component.name }}'");
 
         // Await results
-        child.wait().expect("failed to install component '{{ component.name }}'");
+        let status = child.wait().expect("Error occurred while waiting to install component '{{ component.name }}'");
+
+
+        if !status.success() {
+            let error = child.stderr.take();
```
bitwalker, line 230: This will always be empty, because we're inheriting `stdout`/`stderr` from the parent process (due to the use of `Stdio::inherit` above).

src/commands/update.rs
```diff
@@ -148,11 +175,39 @@ fn update_channel(
...
                     .with_context(|| format!("failed to uninstall {crate_name} via cargo"))?;
 
-                remove_exe
+                let status = remove_exe
                     .wait()
-                    .context("failed to uninstall component '{{ component.name }}'")?;
+                    .context(format!("Error occurred while waiting to uninstall {crate_name}",))?;
+
+                if !status.success() {
+                    let error = remove_exe.stderr.take();
```
bitwalker, line 183: Similar to the other comment, this will always be empty, because of we're inheriting the parent process descriptors via `Stdio::inherit`, so the error information will already be printed to the output.

src/commands/update.rs
```diff
@@ -133,9 +136,33 @@ fn update_channel(
...
+
+                    let error_msg = if let Some(mut error) = error {
+                        let mut stderr_msg = String::new();
+                        let read_err_msg = error.read_to_string(&mut stderr_msg);
+
+                        if read_err_msg.is_err() {
+                            String::from("")
```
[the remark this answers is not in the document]
lima-limon-inc: Noted! Changed here: https://github.com/0xMiden/midenup/pull/41/commits/14574bd17aeec4f6eaa684e362084867a9a545f2

src/commands/install.rs
lima-limon-inc: I see, thanks for the explanation! My main intention was for the error message to get displayed. So with that in mind I think it's best to remove the check entirely. Done here: https://github.com/0xMiden/midenup/commit/dfff669316f24d5d82e51e07ea2607f717153501

src/commands/update.rs
lima-limon-inc: Also changed here: https://github.com/0xMiden/midenup/commit/dfff669316f24d5d82e51e07ea2607f717153501

PR #51, opened 2025-07-16
refactor: separate `commands::init` into two separate functions

### LINKED ISSUE #50: refactor: separate `init` into two: setup `midenup` dir and install stable toolchain
Separate the init function into two:
- One function handles `midenup` directory creation
- The other function handles the other "side effects" that come after init is called. Currently this is:
    - Installing the `stable` toolchain

lima-limon-inc:
Closes #50

Separates the `commands::init` function into two separate parts.
The main bulk of functionality was moved to the `config::ensure_midenup_home_exists` function. This was done to homogenize guarantee that the midenup directory always initialized consistently.

Now, the `commands::init` function also installs stable if it is not present in the midenup directoy. This was the original intended behavior that was missing.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -1,27 +1,18 @@
-use anyhow::Context;
-
-use crate::{commands, config::ensure_midenup_home_exists, manifest::Manifest, Config};
+use crate::{config::ensure_midenup_home_exists, manifest::Manifest, Config};
 
 /// This is the first command the user runs after first installing the midenup. It performs the
 /// following tasks:
 ///
 /// - Bootstrap the `midenup` environment (create directories, default config, etc.), if not already
 ///   done. (done in [ensure_midenup_home_exists]).
-/// - Install the stable channel.
-pub fn init(config: &Config, local_manifest: &mut Manifest) -> anyhow::Result<()> {
+pub fn init(config: &Config, _local_manifest: &mut Manifest) -> anyhow::Result<()> {
     ensure_midenup_home_exists(config)?;
 
-    let toolchains_dir = config.midenup_home.join("toolchains");
-    let stable = toolchains_dir.join("stable");
-    if !stable.exists() {
-        std::println!("About to install stable toolchain");
-
-        let upstream_stable = config
-            .manifest
-            .get_latest_stable()
-            .context("ERROR: No stable channel found in upstream")?;
-        commands::install(config, upstream_stable, local_manifest)?;
-    }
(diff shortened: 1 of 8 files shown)


src/commands/init.rs
```diff
@@ -1,73 +1,27 @@
...
-
-    // Write the symlink for `miden` to $MIDENUP_HOME/bin
-    let current_exe =
-        std::env::current_exe().expect("unable to get location of current executable");
-    let miden_exe = bin_dir.join("miden");
-    if !miden_exe.exists() {
-        utils::symlink(&miden_exe, &current_exe)?;
-    }
+///   done. (done in [ensure_midenup_home_exists]).
+/// - Install the stable channel.
```
bitwalker, line 10: I don't know if we want to make this part of `init`. The `init` command is essentially a part of the installation process for `midenup` itself, rather than anything to do with toolchain installation per se.

Perhaps a better way to handle `init` is to conditionally perform it when installing _any_ toolchain for the first time (which in most cases would be stable, but not necessarily). If `init` has already been performed, then it is a no-op, otherwise it takes care of setting up the `midenup` home directory so it is available for subsequent toolchain operations.

bitwalker (changes requested):
Looks good overall, but I think I would prefer to revamp how we handle `init`, as described in a comment I left above, so that installing `stable` isn't mixed up with initialization of `midenup`.

src/commands/init.rs
lima-limon-inc: > I don't know if we want to make this part of `init`. The `init` command is essentially a part of the installation process for `midenup` itself, rather than anything to do with toolchain installation per se.

I completely agree.

> Perhaps a better way to handle `init` is to conditionally perform it when installing _any_ toolchain for the first time (which in most cases would be stable, but not necessarily). If `init` has already been performed, then it is a no-op, otherwise it takes care of setting up the `midenup` home directory so it is available for subsequent toolchain operations.

Actually, the idea behind `fn ensure_midenup_home_exists()` was for it to perform this very logic. However, considering that `init` does not install stable by default; I think it makes sense to move `ensure_midenup_home_exists`'s logic over to `init` entirely. 

I'll push these changes ASAP.

lima-limon-inc: Updated here: https://github.com/0xMiden/midenup/pull/51/commits/1718497f5d49f547004ef4f807b549f9caa67c4f

bitwalker (approved):

PR #62, opened 2025-07-23
feat: `midenup override`

### LINKED ISSUE #60: feat: include a mechanism to define a default toolchain
### Feature description

There should be a way for users to define a "default" toolchain, besides the project-specific setting done with `midenup set`.
This would be somewhat analogous to `rustup override`.

### Why is this feature needed?

This would be a more ergonomic way to switch between toolchains.

lima-limon-inc:
Closes #60 

This PR implements the `midenup override`; which sets a default global Toolchain; similar to `rustup override`.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1,6 +1,6 @@
 mod init;
 mod install;
-mod override_command;
+mod r#override;
 mod set;
 mod show;
 mod uninstall;
@@ -9,7 +9,7 @@ mod update;
 pub use self::{
     init::init,
     install::{INSTALLABLE_COMPONENTS, install},
-    override_command::override_command,
+    r#override::r#override,
     set::set,
     show::ShowCommand,
     uninstall::uninstall,
--- a/src/commands/override.rs
+++ b/src/commands/override.rs
@@ -1,5 +1,4 @@
-// This function is called override_command because "override" is a reserved
-// keyword.
+// This function is called r#override because "override" is a reserved keyword.
 // Source: https://doc.rust-lang.org/reference/keywords.html#r-lex.keywords.reserved
 
 use anyhow::Context;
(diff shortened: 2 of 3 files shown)


src/commands/mod.rs
```diff
@@ -1,5 +1,6 @@
 mod init;
 mod install;
+mod override_command;
```
bitwalker, line 3: You can do the following instead here so we can keep our naming convention:

```rust
mod r#override;

pub use self::r#override::r#override;
```

The `override` function itself can be defined like:

```rust
pub fn r#override(...) -> anyhow::Result<()> {
    // ...
}
```

bitwalker (changes requested):
LGTM! Just one change I'd like to make and we can merge this.

src/commands/mod.rs
lima-limon-inc: Nice! Thanks for the tip.
Changed here: https://github.com/0xMiden/midenup/pull/62/commits/05ba66180cb810ec4516569caf3b192d3b6b53f8

src/toolchain.rs
```diff
@@ -57,34 +57,69 @@ impl Toolchain {
...
-    pub fn current() -> anyhow::Result<Self> {
-        let toolchain_file = Self::toolchain_file()?;
-        if !toolchain_file.exists() {
-            // The default toolchain is stable
-            //
-            // TODO(pauls): If we support setting global defaults at some point, we'll need
-            // to adjust this.
-            return Ok(Self::default());
+    /// Returns the current active Toolchain according to the following prescedence:
+    /// 1. The toolchain specified by a `miden-toolchain.toml` file in the present working directory
```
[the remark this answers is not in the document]
bitwalker: Sounds good!

bitwalker (approved):

PR #68, opened 2025-08-01
fix: replace the call to `miden --version` with a call to `miden help`

### LINKED ISSUE #66: bug: `midenup` gets stuck after PATH is exported
When `midenup` runs the `miden` command to check if it is accesible, it gets stuck.

lima-limon-inc:
Closes #66 

If `miden <COMPONENT>` gets run  and the `miden` executable *IS* present in the PATH, `miden` gets stuck. This happens in the call to `std::process::Command (...) `miden --version`. `miden` does not recognize `--version` as a flag, so it got replaced with `help`.

With this patch, `miden` no longer gets stuck.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -78,7 +78,7 @@ pub fn setup_midenup(config: &Config) -> anyhow::Result<bool> {
         .stdout(std::process::Stdio::null())
         .stderr(std::process::Stdio::null())
         .stdin(std::process::Stdio::null())
-        .arg("--version")
+        .arg("help")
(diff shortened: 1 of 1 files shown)


bitwalker:
Oh, we should add support for `--version`, and have it output the actual Cargo version of `midenup` as well as the git revision it was built with. We'll want that for troubleshooting user issues.

bitwalker:
I'll merge this for now, but can you add an issue to implement the `--version` handling described above and switch back to using `--version` once that is done?

PR #54, opened 2025-07-16
ci: parallelize integration tests

lima-limon-inc:
Closes #36 
Closes #79 
Closes #91 

This PR aims to parallelize the integration tests. 

This was achieved by installing the toolchains found on the manifest before hand, and then running the tests. 
It would appear that `rustup`'s concurrency problems arise when multiple toolchains are trying to get installed at the same time. 
Admittedly, this conclusion is purely experimental; however it does appear to follow that if the toolchains are installed before hand, there are no problems with concurrent usage.

Additionally, the integrations tests now install the components in `dev` mode which also speeds test execution up.

Since speeding installation up is also useful for development, the `debug` "flag" was made part of the `Config` struct. However, it is only configurable via the `MIDENUP_DEBUG_MODE` environment variable which is hidden.

PR #73, opened 2025-08-04
feat: aliases in manifest

lima-limon-inc:
Close https://github.com/0xMiden/midenup/issues/79 (Cherry picked from https://github.com/0xMiden/midenup/pull/54)

This PR removes the hardcoded aliases from the `midenup` executables and moves them to the Manifest. A new value  was added to the `Component` struct called `aliases` which represents all the aliases that are associated with that component. These aliases are in the form of a map, that goes from the name of the alias to the associated list of commands.

Additionally, now the toolchain's `bin/` holds a series of symlinks. 
Firstly, for each known component, a symlink of the form `miden <component name>` is created. Like so:
```
midenup
├── bin
│   └── miden -> /Users/fabri/Repositories/midenup/miden
├── manifest.json
└── toolchains
    └── 0.15.0
        ├── bin
        (...)
        │   ├── cargo-miden
        │   ├── miden
        │   ├── miden cargo-miden -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/cargo-miden
        │   ├── miden client -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden-client
        │   ├── miden midenc -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/midenc
        │   ├── miden vm -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden
        │   ├── miden-client
        │   ├── midenc
        (...)
```
This is done to "trick" clap into providing `midenup` compatible help messages. Now, when `miden client` is run, we execute the `miden client` symlink, instead of the `miden-client` executable itself. This produces the following output. 
```
CLI actions

Usage: miden client [OPTIONS] <COMMAND>

Commands:
(...)
```
Instead of the usual:
```
CLI actions

Usage: miden-client [OPTIONS] <COMMAND>

Commands:
(...)
```

Additionally, a symlink is created for each known alias (simulate, account, faucet, etc), like mentioned in this comment (https://github.com/0xMiden/midenup/pull/43#issuecomment-3124551070). These point to the symlink which points to the corresponding binary, like so:
```
midenup
├── bin
│   └── miden -> /Users/fabri/Repositories/midenup/miden
├── manifest.json
└── toolchains
    └── 0.15.0
        ├── bin
        │   ├── account -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden client
        │   ├── build -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden cargo-miden
        │   ├── call -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden client
        │   ├── cargo-miden
        │   ├── deploy -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden client
        │   ├── miden
        │   ├── miden cargo-miden -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/cargo-miden
        │   ├── miden client -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden-client
        │   ├── miden midenc -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/midenc
        │   ├── miden vm -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden
        │   ├── miden-client
        │   ├── midenc
        │   ├── send -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden client
        │   └── simulate -> /Users/fabri/Library/Application Support/midenup/toolchains/0.15.0/bin/miden client
```

However, these aliased symlinks are *not* passed to `std::Command` as is. This is because, when used, clap uses the alias's name to display help messages:
```
$ miden account
error: the following required arguments were not provided:
  --account-type <ACCOUNT_TYPE>

Usage: account new-account --account-type <ACCOUNT_TYPE>
```
The main problem being that `account new-account` is not a valid command. 

So, we resolve the alias and use the component's executable instead (i.e `miden account` executes `miden client new-account`). With this, a valid help message is displayed:

```
error: the following required arguments were not provided:
  --account-type <ACCOUNT_TYPE>

Usage: miden client new-account --account-type <ACCOUNT_TYPE>
```

### DIFF
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -29,7 +29,15 @@
           "repository_url": "https://github.com/0xMiden/miden-client.git",
           "crate_name": "miden-client-cli",
           "revision": "83df2aa115b2617e211d1d929a1189ffabbd137b",
-          "installed_executable": "miden-client"
+          "installed_executable": "miden-client",
+          "aliases": {
+              "account": [{"resolve": "client"}, {"verbatim": "new-account"}],
+              "faucet": [{"resolve": "client"}, {"verbatim": "mint"}],
+              "deploy": [{"resolve": "client"}, {"verbatim": "new-wallet"}, {"verbatim": "--deploy"}],
+              "call": [{"resolve": "client"}, {"verbatim": "call"}, {"verbatim": "--show"}],
+              "send": [{"resolve": "client"}, {"verbatim": "send"}],
+              "simulate": [{"resolve": "client"}, {"verbatim": "exec"}]
+            }
         },
         {
           "name": "midenc",
@@ -39,7 +47,11 @@
         {
           "name": "cargo-miden",
           "version": "0.1.0",
-          "rustup_channel": "nightly-2025-03-20"
+          "rustup_channel": "nightly-2025-03-20",
+          "aliases": {
+              "new": [{"verbatim": "cargo"}, {"verbatim": "miden"}, {"verbatim": "new"}],
+              "build": [{"verbatim": "cargo"}, {"verbatim": "miden"}, {"verbatim": "build"}]
+            }
         }
       ]
     },
@@ -69,7 +81,15 @@
           "name": "client",
           "package": "miden-client-cli",
           "version": "0.10.0",
-          "installed_executable": "miden-client"
+          "installed_executable": "miden-client",
+          "aliases": {
+              "account": [{"resolve": "client"}, {"verbatim": "new-account"}],
+              "faucet": [{"resolve": "client"}, {"verbatim": "mint"}],
+              "deploy": [{"resolve": "client"}, {"verbatim": "new-wallet"}, {"verbatim": "--deploy"}],
+              "call": [{"resolve": "client"}, {"verbatim": "call"}, {"verbatim": "--show"}],
+              "send": [{"resolve": "client"}, {"verbatim": "send"}],
+              "simulate": [{"resolve": "client"}, {"verbatim": "exec"}]
+            }
         },
         {
           "name": "midenc",
@@ -79,7 +99,11 @@
         {
           "name": "cargo-miden",
           "version": "0.1.0",
-          "rustup_channel": "nightly-2025-03-20"
+          "rustup_channel": "nightly-2025-03-20",
+          "aliases": {
+              "new": [{"verbatim": "cargo"}, {"verbatim": "miden"}, {"verbatim": "new"}],
+              "build": [{"verbatim": "cargo"}, {"verbatim": "miden"}, {"verbatim": "build"}]
+            }
         }
       ]
     }
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -80,7 +80,7 @@
         {
           "name": "client",
           "package": "miden-client-cli",
-          "version": "0.10.0",
+          "version": "0.10.2",
           "installed_executable": "miden-client",
           "aliases": {
               "account": [{"resolve": "client"}, {"verbatim": "new-account"}],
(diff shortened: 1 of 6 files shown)


src/miden_wrapper.rs
```diff
@@ -316,92 +256,49 @@ miden help"
...
-            // NOTE: This value was added in order to both:
-            // - Emulate clap's padding
-            // - Improve readability
-            let padding = 3;
-            let spaces = String::from(' ').repeat(spacing + padding);
-            format!("  {}{}{}\n", alias.bold(), spaces, description)
-        })
-        .collect();
-
+    let asterisk = "*".bold();
```
lima-limon-inc, line 291: Since now aliases are also obtained from the manifest, the `miden help` message can not display them without possibly issuing an install. Now the `miden help toolchain` command  displays both the aliases and the components.

manifest/channel-manifest.json
```diff
@@ -69,7 +81,15 @@
...
-          "installed_executable": "miden-client"
+          "installed_executable": "miden-client",
+          "aliases": {
+              "account": [{"resolve": "client"}, {"verbatim": "new-account"}],
+              "faucet": [{"resolve": "client"}, {"verbatim": "mint"}],
+              "deploy": [{"resolve": "client"}, {"verbatim": "new-wallet"}, {"verbatim": "--deploy"}],
+              "call": [{"resolve": "client"}, {"verbatim": "call"}, {"verbatim": "--show"}],
+              "send": [{"resolve": "client"}, {"verbatim": "send"}],
+              "simulate": [{"resolve": "client"}, {"verbatim": "exec"}]
+            }
```
lima-limon-inc, line 92: I added this distinction between "resolve" and "verbatim" because of the way `cargo-miden`'s aliases are resolved.

Originally, I thought of simply having a list of known arguments to pass to the component's corresponding binary, like so:
```
          "name": "client",
          (...)
          "aliases": {
              "account": ["new-account"],
              "deploy": ["new-wallet", "--deploy"],
          (...)
          }
```
However,  `cargo-miden`'s aliases require adding parameters before *and* after the executable. For instance:
`miden new dir` -> `cargo miden new dir`
 
So, in order to represent them both, I went with this "generic" approach, where the alias resolution spells out the entirety of the command to run, and denotes which parameters need to be resolved at runtime.

Currently, arguments marked as "resolved" are resolved via the `installed_file` field; although this could potentially be generalized.

lima-limon-inc:
Heads up: I'll rebase this PR

fkrause98 (approved):
Tested locally and works as expected, let's address the comment I left about the documentation, approving to avoid blocking this.

manifest/channel-manifest.json
lima-limon-inc: > It's good that you've specified your reasoning here, but I'd like to see this kept in a doc of some form, can we add this to an existing doc or create one to explain this? Even a comment in the code can work, but let's keep this somewhere we can read about it, this PR comment will eventually get lost.

Noted, I added an explanation in this commit: https://github.com/0xMiden/midenup/pull/73/commits/6ae9dff15872deb1912123e7cf653fa77098a7ad

src/commands/install.rs
```diff
@@ -291,6 +315,36 @@ fn main() {
...
+            if let InstalledFile::Executable { ref binary_name } = exe_name {
+                let miden_prefix = format!("miden {}", component.name);
+                for alias in aliases {
+                    acc.insert(alias.clone(), miden_prefix.clone());
+                }
+                acc.insert(miden_prefix, binary_name.clone());
+            }
+            acc
+        })
+        .iter()
```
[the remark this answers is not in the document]
lima-limon-inc: Great point, changed here: https://github.com/0xMiden/midenup/pull/73/commits/02393dd74dc1716e4d34efe69fd84411cb1cf3f1

PR #77, opened 2025-08-08
feat: advertise component initialization in Manifest

lima-limon-inc:
Currently, all aliased commands that map to a component that requires initialization will fail, unless the user knows to initialize them ahead of time.  

This PR introduces an `initialization` field to the `Component` struct, by which components advertise the fact that they require initialization in the Manifest. 

At the moment, this is merely shown on the `miden help toolchain` message, since initialization does require user intervention (by means of passing arguments).


lima-limon-inc:
Sidenote: Also rebased to the latest changes on `main`.

Issue #78, opened 2025-08-11
feat: implement `--version` flag for `miden`.


lima-limon-inc:
`miden` should support a `--version` that outputs both the Cargo version of midenup as well as the git revision it was built with.

Once implemented, [this call to `miden help`](https://github.com/0xMiden/midenup/blob/5a37665eca6235ae4d62763fe7e912ede33016fe/src/commands/init.rs#L81) should be replaced with a call to `miden --version`.

PR #81, opened 2025-08-11
feat: add `--version` flag support for `miden`

lima-limon-inc:
Closes: #78 

This PR implements the `--version` flag for the `miden` command. This displays:
- The system's cargo version
- Midenup's version
- The rev-hash midenup was compiled under (this is obtained at compile time by setting the `GIT_REV` env variable in the Makefile).
- The cargo version Midenup was compiled with (as a bonus)

Additionally, "create issue" message is displayed. When pasted onto the browser, it will automatically create an issue with that information displayed (abbreviated for clarity).


build.rs
```diff
@@ -0,0 +1,39 @@
+use std::process::Command;
+
+fn main() {
+    // We delete the old build/ directory in order to avoid mixing files up.
+    if std::path::PathBuf::from("build").exists() {
```
bitwalker, line 5: We should put these files in `$OUT_DIR`, a directory created by Cargo when running build scripts - it can be obtained using `env!("OUT_DIR")`.

Then, from within source files, you can do things like `include_str!(concat!(env!("OUT_DIR"), "git_version.in"))` or whatever to read the files you placed there.

bitwalker (changes requested):
Like the output! One thing I'd like to change about how we're doing the build related stuff, but otherwise looks good.

PR #87, opened 2025-08-19
[DNM] demo: demo branch with everything merged

lima-limon-inc:
This is a demonstration branch, which contains every feature included in all the opened PRs.

This is not meant to be merged.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1,0 +1,199 @@
+# Installing midenup
+Firstly, midenup can be installed with the following command:
+
+```shell
+cargo install midenup
+```
+
+> [!IMPORTANT]
+> Until this crate has been published to crates.io, it is only possible to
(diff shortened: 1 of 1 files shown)


bitwalker:
With all of the non-draft PRs merged, where do we sit with this branch? After merging #97, I'm seeing CI failures as well, so we should at least make sure those are addressed.

lima-limon-inc:
> With all of the non-draft PRs merged, where do we sit with this branch? 

I believe we can safely delete this branch; all features it included have been merged IINM. 

> After merging #97, I'm seeing CI failures as well, so we should at least make sure those are addressed.

Noted, I've created https://github.com/0xMiden/midenup/pull/103 to address it.

bitwalker:
Great! Feel free to close this once you feel it is no longer needed

PR #89, opened 2025-08-20
feat: miden only installs components listed in `miden-toolchain.toml` file

### LINKED ISSUE #88: feat: adding components to `miden-toolchain.toml` after toolchain installation
Issue: 
An empty component list in a `miden-toolchain.toml` file does not get updated after the toolchain gets installed.

Context:
Currently, if `midenup set <toolchain>` command is run, a `miden-toolchain.toml` file gets generated which contains:
- The toolchain version number
- The list of components available in that version

If the toolchain is not installed when the command is run, the component list is left empty and the following message gets displayed:

```
WARNING: Non present toolchain was set. Component list will be left empty
```

However, this list does not get updated after that toolchain has been installed; and is thus left outdated.

lima-limon-inc:
Closes #88 
Depends on #114

Originally, the component list present in miden-toolchain.toml was used to reflect the available components present upstream. However, with these new changes, miden will only install the components explicitly listed on the miden-toolchain.toml file.

The behavior is now as following:
- If a user runs `midenup install <channel>`, the entire channel will be installed. 
- If `miden` detects a `miden-toolchain.toml` file, it will only install the components present in the `[component]` table
    - If the list is empty, the entire channel is installed.
- `miden` will only display components/libraries present in the current active toolchain in commands such as `miden help toolchain`.
    - However, if the user attempts to use a component that _is_ installed in the system, but it is *not* present in the active toolchain, `miden` will execute the component regardless (after having displayed a warning message).

Sidenote: Since the `.masp` libraries are now optional, they are now marked as dependencies for the compiler (via the 
 `required` attribute).

Let me know your thoughts! Thanks in advance 😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,10 +175,24 @@ For example, to set `0.16.0` run:
 midenup set 0.16.0
 ```
 
-Now, whenever `miden` is called in this directory, it will use the specified toolchain.
+This procedure will generate a `miden-toolchain.toml` file in the directory where `midenup set` was invoked:
 
-> [!NOTE]
-> This procedure generates a `miden-toolchain.toml` file in the directory where `midenup set` was invoked.
+```toml
+[toolchain]
+channel = "stable"
+components = []
+```
+
+Now, whenever `miden` is called in this directory (or any of its subdirectories), it will use the specified toolchain.
+If the `components` entry is left blank, all the available components for the selected channel will be installed. However, if the list is not empty, only the listed components will be installed.
+For example, with the following `miden-toolchain.toml` file:
+```toml
+[toolchain]
+channel = "stable"
+components = ["vm", "midenc", "client"]
+```
+Only the `vm`, `midenc`, `client` will be installed after `miden` gets executed.
+
 
 #### Setting a global default toolchain
 


src/channel.rs
```diff
@@ -130,6 +178,90 @@ impl Channel {
...
+            println!();
+            // TODO: Add messages for the other justifications
+            #[allow(clippy::single_match)]
+            match toolchain_justification {
+                ToolchainJustification::MidenToolchainFile { path } => println!(
+                    "Check the `miden_toolchain.toml` in {} to see if any \
+                         component is misspelled or got removed from upstream",
+                    path.display()
+                ),
+                _ => (),
```
lima-limon-inc, line 253: This produces the following output:
```
WARNING: Some elements present in the current Toolchain are not present in the upstream channel: 0.16.0

- hocus, which was explictely selected for installation, is missing in upstream channel
- bogus, which was explictely selected for installation, is missing in upstream channel
- pocus, which was explictely selected for installation, is missing in upstream channel

These components will be ignored for the current install.

Check the `miden_toolchain.toml` in /Users/fabri/Repositories/midenup/miden-toolchain.toml to see if any component is misspelled or got removed from upstream

```

bitwalker (approved):

PR #112, opened 2025-10-16
feat: add miden-node to components list + automatic initialization

### LINKED ISSUE #110: Add a command for spinning up a node locally
# Summary

This issue is a result of the [`midenup` Improvement Proposal discussion](https://github.com/0xMiden/midenup/discussions/108).

## Description

Adding a command for running a local node would greatly improve dev experience. This would be super useful for devs that want to run things locally.

### Input command

```bash
miden node
```

lima-limon-inc:
Closes #110 

Depends on #114

This PR adds the Miden node to `midenup`'s manifest.

Additionally, it adds automatic initialization functionality to all components that require it. After installing a component advertises initialization, `midenup` will execute said commands to automatically initialize it.

In the case of the node, it will use the `toolchains/<version>/var/accounts/` and `toolchains/<version>/var/data/` directories as its default saving spot.

### DIFF
--- a/tests/data/integration_install_from_non_cargo/channel-manifest-1.json
+++ b/tests/data/integration_install_from_non_cargo/channel-manifest-1.json
@@ -40,17 +40,6 @@
               "send": ["executable", "send"],
               "simulate": ["executable", "exec"]
            }
-        },
-        {
-          "name": "midenc",
-          "version": "0.1.0",
-          "requires": ["base", "std"],
-          "rustup_channel": "nightly-2025-03-20"
-        },
-        {
-          "name": "cargo-miden",
-          "version": "0.1.0",
-          "rustup_channel": "nightly-2025-03-20"
         }
       ]
     }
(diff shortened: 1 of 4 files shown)


Keinberger:
> > If I understand correctly, this command not only adds the `miden node` alias, but also adds automatic initialization of the `miden client`, is that correct?
> 
> Correct! It does not _explicitely_ auto initialize the client, but rather attempts to initialize every component which advertises an initialization command.
> 
> > Since I opened this PR to add silent initialization to the `miden client` when running a client command, initializing the `miden-client` would no longer be required. Given that, not sure if we wanna keep this here.
> 
> I belive the previous releases of the client would still require manual initialization. What we could do, once your PR gets merged, is to simply remove the `initialization` field from the client's `0.12.0` entry in the manifest.
> 
> With that, `midenup` would manually auto initialize the client for the previous releases and would delegate auto-initialization for release `0.12.0` (and potentially onward).
> 
> Do is it make sense?

That's a great idea, let's do that!

lima-limon-inc:
> Seems to me like maintaining an initialization command list for components that already support it or are about to support it might be unnecessary overhead. I would consider removing the initialization logic from `midenup` and delegating to components instead.
> 
> EDIT: I thought having hte command in the manifest was a new addition, but seems like it was not and the only addition in this PR was automatically running it. Still, not sure I'd add this but not a strong opinion.

I think I agree with @igamigo on this. I believe removing that field would remove some potentially unnecessary complexity, especially considering these latest batch of changes.

PR #117, opened 2025-10-23
fix: update toolchain with new components

### LINKED ISSUE #116: bug: `midenup update` panics if a toolchain has a new component
Currently, midenup tries to remove the pre-existing component files when performing an update indiscriminately. 

However, if a toolchain version had a new component added to it, then there's nothing to delete; but midenup tries to do so anyway.

Before deleting the component, it should check if there's anything to delete.

lima-limon-inc:
Closes #116 

Depends on: #114 

If an update introduced a new component in the toolchain, there are no files that require deletion. However, `midenup` was not checking what de cause of the update was and was deleting old files indiscriminately (even when there were no old files). 
This PR makes the `components_to_update` function return a new value called `UpdateMotive` which indicates in what way a component must be updated.

### DIFF
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -344,20 +344,23 @@ pub fn components_to_update(older: &Channel, newer: &Channel) -> Vec<(Component,
 
     // These are the elements that are present in boths sets. We are only
     // interested in those which need updating.
-    let components_to_update = current.intersection(&new_channel).filter_map(|current_component| {
-        let new_component = new_channel.get(*current_component);
-        match new_component {
-            // This should't be possible, but if somehow the component is
-            // missing, then we trigger an update for said component regardless.
-            None => Some((current_component, UpdateMotive::Added)),
-            // We only want to update components that share the same name but
-            // differ in some other field.
-            Some(new_component) if !current_component.is_up_to_date(new_component) => {
-                Some((current_component, UpdateMotive::NewerVersion))
-            },
-            _ => None,
-        }
-    });
+    let components_to_update = current
+        .iter()
+        .filter(|comp| new_channel.contains(**comp))
+        .filter_map(|current_component| {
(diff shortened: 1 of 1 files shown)


lima-limon-inc:
> There appears to be a bug in this PR which got caught in `integration_update_test`. I'm still debugging it.
> 
> When `midenup update` is called, the update process fails to detect that the `miden-vm` present in version 0.15.0 of the toolchain got a version bump. It _does_ recognize that the `bast.masp` library must be removed however.
> 
> Additionally, it would appear as if a second call to `midenup update` does detect this bump and updates the vm accordingly.
> 
> The intended behavior is for both components to get updated in the same call. I'll update this PR as soon as I find a fix.

I think the problem is in the intersection method in `HashSet`: https://doc.rust-lang.org/stable/std/collections/struct.HashSet.html#method.intersection

It states:
> may yield references to one or the other. This can be relevant if T contains fields which are not compared by its Eq implementation, and may hold different value between the two equal copies of T in the two sets

This affects our approach since `Eq` for `Component` only checks for the name, not the rest of the fields. For whatever reason, it is only now returning `other` instead of `self`. I'll change the implementation so that it is deterministic.

lima-limon-inc:
> > may yield references to one or the other. This can be relevant if T contains fields which are not compared by its Eq implementation, and may hold different value between the two equal copies of T in the two sets
> 
> This affects our approach since `Eq` for `Component` only checks for the name, not the rest of the fields. For whatever reason, it is only now returning `other` instead of `self`. I'll change the implementation so that it is deterministic.

Update: it appears that the bug was indeed caused by `HashSet::intersection()`, this commit seems to have fixed it: https://github.com/0xMiden/midenup/pull/117/commits/ea937b43beb3cd32aaea6f22f5ec0b5765633af6

Issue #121, opened 2025-11-03
feat: add artifact support


lima-limon-inc:
Currently, `midenup` compiles every component from source when performing an install.

Artifact support would enable `midenup` to install a pre-built binary of a specific component; which would speed install times significantly. 

If a component does not have a pre-built binary available for the target architecture or were the download of said file to fail; then `midenup` would use its current behavior as a fallback.

This will probably also require expanding the `manifest.json` in order to advertise said binaries.

PR #122, opened 2025-11-03
feat: artifact support

lima-limon-inc:
Closes: #121 

This PR adds artifact support, both for executables and libraries. With these changes, `midenup` should first try to download an artifact if it exists, if it doesn't or the download process fails, then it defaults to install the component from source.  

Implementation notes:
- The `Artifact` struct holds a URI which is used to fetch the artifact, either from a URL (prefixed by `https://`) or from a file (prefixed by `file://`). Each artifact only holds a specific triplet (for more information, see [LLVM's triplet implementation](https://llvm.org/doxygen/Triple_8h_source.html)), which is displayed in the URI itself, for example:
    - `https://github.com/0xMiden/miden-debug/releases/download/v0.4.4/miden-debug-aarch64-apple-darwin`: : Holds the `miden-debugger`'s binary for `aarch64-apple-darwin`
    - `https://github.com/0xMiden/miden-vm/releases/download/0.0/std.masp`: Holds the `Package` for the `std` library. (Do note that `masp` packages are used by the MidenVM and therefore these URIs do not have a platform dependent triplet). 
- The `src/external.rs` file holds functions that are imported into the install script by  the `include_str` macro. These were added in a separate file to enhance readability. 
- As a final note, now if a component fails to get installed, midenup will simply skip said component if it's in release mode. In debug mode, it panics.

### DIFF
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -271,7 +271,7 @@
           }
         },
         {
-          "name": "debugger",
+          "name": "debug",
           "package": "miden-debug",
           "version": "0.4.4",
           "installed_executable": "miden-debug",


src/commands/install.rs
```diff
@@ -37,7 +37,31 @@ pub fn install(
...
+
+    // opt/ directory which holds symlinks to binaries in bin/.
+    // These are used in order to preserve a "midenup" compatible
+    // interface. This relies on the fact that clap uses argv[0] in order to
+    // display executable names names. These symlinks have the following format:
+    // `miden <component name>`
+    // Then, when `miden` is invoked, it uses these symlinks to execute the
+    // underlying binary.  With this setup, `clap` displays the name as: `miden
+    // <component name>` instead of just `binary_name` when displaying help
+    // messages.
```
lima-limon-inc: Neither `opt/`, `bin/` nor `lib/` changed in this PR. I simply added these lines and comment to elaborate on their functionality.

manifest/channel-manifest.json
```diff
@@ -123,7 +123,13 @@
           "package": "miden-stdlib",
           "version": "0.17.2",
           "installed_library": "std.masp",
-          "library_struct": "miden_stdlib::StdLibrary"
+          "library_struct": "miden_stdlib::StdLibrary",
+          "artifacts": [
+              {
+                  "target": "zkvm-miden-unknown",
+                  "uri": "https://github.com/lima-limon-inc/fabriz.io/releases/download/0.0/std.masp"
```
lima-limon-inc: These URLS are not intended to be merged and are only here to test this PR. They should be removed before merging.

lima-limon-inc:
Sidenote: I'll give the implementation a rework. I'll replace the triplet to url table with a list of urls instead to simplify things.

bitwalker:
@lima-limon-inc I finally got the the release workflow for `miden-debug` sorted with artifact (with attestation) support working, with the [v0.4.4](https://github.com/0xMiden/miden-debug/releases/tag/v0.4.4) release.

v0.4.4 can be used with the v0.19 toolchain, upcoming releases will work with v0.20+. In the meantime, hopefully that provides all that is needed to validate a real component works with this feature.

lima-limon-inc:
> @lima-limon-inc I finally got the the release workflow for `miden-debug` sorted with artifact (with attestation) support working, with the [v0.4.4](https://github.com/0xMiden/miden-debug/releases/tag/v0.4.4) release.
> 
> v0.4.4 can be used with the v0.19 toolchain, upcoming releases will work with v0.20+. In the meantime, hopefully that provides all that is needed to validate a real component works with this feature.

@bitwalker awesome, thanks for the ping! 

I have just added the debugger in the [0.19.0 toolchain](https://github.com/0xMiden/midenup/pull/122/changes#diff-b147fa66f2f576172bd50fd74ea73cff00c8a2dae2ca577dea3f41e623740e3dR272-R280) with its corresponding artifact field in this commit: https://github.com/0xMiden/midenup/pull/122/commits/a85fb790d80a6c5bdafa954d3b18ddb7ae7bd61c. 

It only required doing a small change to the way URI's were parsed, here: https://github.com/0xMiden/midenup/pull/122/commits/4d295a218323e6afcdafbca07d7aa301977b1677. Since the link did not contain the component name (i.e. "debugger"), but rather the repository name (i.e. "miden-debug"). However, this ended up simplifying the implementation all together and actually made the URI's more versatile, since now it can be an arbitrary string as long as the triplet is respected.
 
I tested on my machine (Apple M3) and the debugger's binary was fetched successfully and I was able to execute it. Here's the output with the corresponding command:

```
MIDENUP_MANIFEST_URI=file://path/to/local/channel-manifest.json MIDENUP_DEBUG_MODE=true ./midenup install stable

Installing: std.masp
     No artifact found. Proceeding to install from source
     Installed!
Installing: base.masp
     No artifact found. Proceeding to install from source
     Installed!
Installing: vm
     No artifact found. Proceeding to install from source
     Installed!
Installing: client
     No artifact found. Proceeding to install from source
     Installed!
Installing: midenc
     No artifact found. Proceeding to install from source
     Installed!
Installing: cargo-miden
     No artifact found. Proceeding to install from source
     Installed!
Installing: node
     No artifact found. Proceeding to install from source
     Installed!
Installing: debugger
     Fetching artifact
     Installed!

MIDENUP_MANIFEST_URI=file://path/to/local/channel-manifest.json ./miden debugger

WARNING: debugger is installed, but it is not part of the current active toolchain.
error: the following required arguments were not provided:
  <FILE>
```

(Sidenote: the warning message stems from https://github.com/0xMiden/midenup/issues/123, which got fixed in https://github.com/0xMiden/midenup/pull/124).

bitwalker:
@lima-limon-inc Nice work!

The next release of the compiler will have artifacts as well, just like `miden-debug`, so that should help with install time considerably.

One thing to note, `miden-debug` is the name of the executable (corresponding to the convention for commands that are surfaced under the `miden` CLI, i.e. we'd expect to run the debugger using `miden debug`). I see that the component name for the debugger you chose was `debugger`, but I think it might be best for component names to stick close to their executable name, when the component is an executable.

This would mean that we'd end up with the following components currently:

* `midenc`
* `cargo-miden`
* `miden-vm`
* `miden-debug`
* `miden-client`
* `miden-faucet` (or whatever the executable ends up being called)

The non-executable component names are a bit trickier, though I think currently there is only the standard library and protocol library that fall in this bucket. For those, I think we'll want to use the convention of naming the component after the package, e.g. `core` (`std` was renamed to `core` in 0.20) and `base` (the name of this package will likely change too).

Anyway, wanted to align on that since it came up.

lima-limon-inc:
@bitwalker thanks ^_^.

> One thing to note, `miden-debug` is the name of the executable (corresponding to the convention for commands that are surfaced under the `miden` CLI, i.e. we'd expect to run the debugger using `miden debug`). I see that the component name for the debugger you chose was `debugger`, but I think it might be best for component names to stick close to their executable name, when the component is an executable.

Noted, I've pushed this commit https://github.com/0xMiden/midenup/pull/122/commits/819ab9ff3be8a1fe3e3feae8120e5e2eae115f56 which changes the component's name to `debug` so that it more closely resembles the underlying binary. 

> The non-executable component names are a bit trickier, though I think currently there is only the standard library and protocol library that fall in this bucket. For those, I think we'll want to use the convention of naming the component after the package, e.g. `core` (`std` was renamed to `core` in 0.20) and `base` (the name of this package will likely change too).

Great, thanks for the heads up!

bitwalker:
I'll do one last once-over of this today, and try to get it merged ASAP!

Issue #126, opened 2025-12-03
feat: automatic generation/updating of the manifest


lima-limon-inc:
Currently, `midenup`'s manifest, which is used to describe toolchains and the various components that make them up, requires manual intervention when a new toolchain is released.

Ideally, there could be a mechanism by which the manifest is automatically updated/generated. 

I imagine this would involve some type of inter-repository interaction between `midenup`'s repository and the other miden repositories which contain the various components. 

One potential solution would be to have a github action that is triggered when a component gets a new release. This action would then modifiy `midenup`'s manifest file, adding its associated CLI to the corresponding toolchain version entry.
I believe that it automatically figuring out in which toolchain a component belongs to is still an open question. My initial guess would be to deduce this from the crate/workspace's `Cargo.toml` versions.

Would love to hear suggestions/feedback!

bitwalker:
I've implemented a small utility tool `update-manifest` as an extension of #201, which can be used to modify/check/format the channel manifest without having to do it by hand. We can extend this further in the future, but I think it solves the biggest pain point at the moment (which is the fragility of editing the file by hand).

PR #130, opened 2025-12-09
feat: support multiple commands and multiple binaries in aliases

### LINKED ISSUE #129: feat: aliases with multiple commands, from multiple binaries
Stems from: https://github.com/0xMiden/miden-faucet/issues/192, which requires midenup to be able to execute a series of commands, in succession, from different binaries.

Since this involves executing commands from _differente_ binaries, maybe this PR's approach can be revisited: https://github.com/0xMiden/midenup/pull/118

lima-limon-inc:
Closes #129 

This PR makes adds the following functionalities:

- `Channel`s can also have `Aliases`
- Aliases can now contain a series of commands instead of just one. 

Both of these features were added in order to support the `miden mint` command (described in more detail here: https://github.com/0xMiden/miden-faucet/issues/192) which requires a succession of command from different binaries to be executed.

To achieve this, the following changes were made:
- `MidenArgument::Alias` now contains a `Vec<CLICommand>` instead of a single `CLICommand`.  Additionally, the `Component` field was removed, in order to support multi-binary aliases.
- `CliCommand::Executable` now requires the following argument to be of type `CliCommand::Verbatim` with the *name of the component* that needs to be executed.
     - This does mean that component specific aliases now require to redundantly add their own name in the aliases, like [so](https://github.com/0xMiden/midenup/pull/130/files#diff-b147fa66f2f576172bd50fd74ea73cff00c8a2dae2ca577dea3f41e623740e3dR36-R42).
     - Additionally, it does mean that a component could _technically_ use a different component's binary in its alias. While not ideal.
- Added `CliCommand::PositionalArgument`, which can be used to pass specific positional argument used in the CLI to the underlying binaries. This mainly intended for multi-command aliases, like so:
```    
      "aliases": {
          "mint": [
              ["executable", "faucet-client", "mint", "-a", 0],
              ["executable", "client", "sync"],
              ["executable", "client", "consume-notes", "-a", 0]
          ]
      }
```
This means that when a user calls the `miden mint` alias, the first argument they pass to `miden mint` will be passed to `faucet-client` and `client` executables.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -759,7 +759,7 @@ impl Component {
     /// Returns the String representation under which midenup calls a component.
     pub fn get_call_format(&self) -> Vec<CliCommand> {
         if self.call_format.is_empty() {
-            vec![CliCommand::Executable]
+            vec![CliCommand::Executable, CliCommand::Verbatim(self.name.to_string())]
         } else {
             self.call_format.clone()
         }


src/channel.rs
```diff
@@ -368,7 +384,14 @@ impl Display for InstalledFile {
...
 pub enum CliCommand {
-    /// Resolve the command to a [[Component]]'s corresponding executable.
+    /// Resolve the command to a [[Component]]'s corresponding executable.  This
+    /// requires the following CliCommand in the manifest to be a
+    /// [[CliCommand::Verbatim]] with the name of the executable to execute.
+    /// For example:
+    ///
+    ///   "aliases": {
+    ///       "account": [["executable", "client", "new-account"]],
+    ///   }
```
lima-limon-inc: I'm not too fond of `Executable` and `VarPath` depending implicitly on the following CliCommand on the manifest. 
I believe these two could be refactored so that all the required arguments are on the same enum variant.
Like so: 
```shell
pub enum CliCommand {
    (...)
    Executable { component_name: String },
    (...)
```
This would change the way `CliCommand` is serialized, but I think encapsulating all the required component into a single variant would a win in the long run.

lima-limon-inc:
Sidenote: Since this changes the way the Manifest is serialized, it is expected for [this test](https://github.com/0xMiden/midenup/actions/runs/20175734271/job/57923282351) to fail; since it uses the published manifest.

src/manifest.rs
```diff
@@ -239,20 +239,6 @@ mod tests {
...
-        let manifest = Manifest::load_from(Manifest::PUBLISHED_MANIFEST_URI)
-            .expect("Failed to parse upstream manifest.");
-
-        let stable = manifest
-            .get_channel(&UserChannel::Stable)
-            .expect("Could not convert UserChannel to internal channel representation");
-
-        assert!(stable.get_component("std").is_some());
-    }
-
```
lima-limon-inc: I removed this test since installing the stable toolchain is already done in `.github/workflows/integration.yml`. 

Additionally, the integration tests are only run if the unit tests pass. So currently, this test is only blocking the integration test suite; without providing additional functionality.

lima-limon-inc:
Hi there @Keinberger !


> FYI, I think it might make sense to compare the two approaches.

Agreed! I looked at your PR and I believe our approaches are more similar than they are dissimilar.

> On PR #132 , I moved the aliases outside of the individual components struct, to a "aliases" struct that sits at the top-level of each channel (which caused significantly more code changes). 

I also added this [channel-wide aliases](https://github.com/0xMiden/midenup/pull/130/changes#diff-5cb8d80357286351f074b2bac08638050f4bb5e278bc9f4160897a895cc525fcR56-R60) in my PR. I believe having aliases which encompass multiple binaries makes these "channel wide aliases" a necessity.  

I think the main difference in approach between both PRs is that https://github.com/0xMiden/midenup/pull/132 removes component specific aliases, whilst this PR keeps them.

> the future mint command will execute commands form both the miden-faucet and miden-client. Keeping the aliases inside of the individual executables would make that less ideal, as you also pointed out in your PR description

I believe keeping the concept of "component specific aliases" is worthwhile. Were the CLI of a component to change, it would only be _that_ component's responsibility to update the corresponding manifest entries. 
This does depend on the approach we take with regards to https://github.com/0xMiden/midenup/issues/126, i.e., how the manifest is generated in the future; but I imagine that, broadly speaking, it will involve delegating to each component how its manifest entry is generated.

On the other, aliases that correspond to multiple binaries, like `miden mint`, will probably still need special treatment (most likely in midenup repository itself). But I do think that this separation will ease manifest maintainability looking forward, since only these multi-binary aliases would be managed by midenup itself.. 
Would love to hear your take on this @Keinberger. Additionally, just so we are on the same page, this distinction would only exist in the code, from the user's perspective, all the aliases function the same, see [this snippet](https://github.com/0xMiden/midenup/pull/130/changes#diff-5cb8d80357286351f074b2bac08638050f4bb5e278bc9f4160897a895cc525fcR130-R141).

> Furthermore, I think we need better coordination on who is working on what for `midenup` related stuff, since we were both working on the same thing😅

Totally, I apologize for any inconvenience this might have caused. 

Having said this, I'll be working on a slight refactor of `CliCommand` (on a separate PR in order for the core functionality to get merged), tackling this comment: https://github.com/0xMiden/midenup/pull/130#discussion_r2615137935. 

I'm not too certain of going with this [table-like approach](https://github.com/0xMiden/midenup/pull/132/changes#diff-b147fa66f2f576172bd50fd74ea73cff00c8a2dae2ca577dea3f41e623740e3dR100-R101), since that might limit `midenup`'s usability with tools that are not part of `miden` ecosystem itself, (things like `cargo` for instance, like in [this, now closed, PR](https://github.com/0xMiden/midenup/pull/118)).

bitwalker:
> I think some of the rationale can be found in the [closed client PR](https://github.com/0xMiden/miden-client/pull/1508#issuecomment-3612655197) (cc @mmagician).

Based on my reading of that comment, it would seem that the inverse could be done instead then, implement `mint` under `miden-faucet`, which internally uses the `miden-client` crate as necessary. Essentially `miden-faucet` ends up being a specialized `miden-client`.

In general though, a key design constraint is that `miden` isn't actually it's own binary, so we can't implement arbitrary functionality in `miden` (and shouldn't). Instead, it's expected that the underlying components provide that behavior, and `miden` just surfaces it as a unified CLI.

If we determine that we really want `miden` to provide this sort of meta scripting capability, then I would argue that we should make the `miden` CLI itself a separate executable from `midenup`, installed as a toolchain component. Most of the stuff we've forced into the channel manifest around aliases and such would just go away, or be built in to `miden` directly, and `midenup` would retain its simplicity and role as toolchain manager.

Either way, I don't think the direction in this PR or in #132 are the way forward. It strikes me that if the need for this kind of capability is more pervasive than just the `mint` command, then we probably need to split `miden` out as described above, so it can just be as arbitrarily complex as we want.

Keinberger:
@lima-limon-inc 

> I believe keeping the concept of "component specific aliases" is worthwhile. Were the CLI of a component to change, it would only be that component's responsibility to update the corresponding manifest entries.
This does depend on the approach we take with regards to https://github.com/0xMiden/midenup/issues/126, i.e., how the manifest is generated in the future; but I imagine that, broadly speaking, it will involve delegating to each component how its manifest entry is generated.
> On the other, aliases that correspond to multiple binaries, like miden mint, will probably still need special treatment (most likely in midenup repository itself). But I do think that this separation will ease manifest maintainability looking forward, since only these multi-binary aliases would be managed by midenup itself..
Would love to hear your take on this @Keinberger. Additionally, just so we are on the same page, this distinction would only exist in the code, from the user's perspective, all the aliases function the same, see [this snippet](https://github.com/0xMiden/midenup/pull/130/changes#diff-5cb8d80357286351f074b2bac08638050f4bb5e278bc9f4160897a895cc525fcR130-R141).

Thanks for explaining, I agree that it is worthwhile to keep the component-centric aliases. Nevertheless, one suggestion that I'd like to make would be to rename the "key" for channel-wide aliases in the manifest file. Otherwise I fear it's confusing as to why aliases are defined channel-wide, and component-specific.

> Totally, I apologize for any inconvenience this might have caused.
> Having said this, I'll be working on a slight refactor of CliCommand (on a separate PR in order for the core functionality to get merged), tackling this comment: https://github.com/0xMiden/midenup/pull/130#discussion_r2615137935.
> I'm not too certain of going with this [table-like approach](https://github.com/0xMiden/midenup/pull/132/changes#diff-b147fa66f2f576172bd50fd74ea73cff00c8a2dae2ca577dea3f41e623740e3dR100-R101), since that might limit midenup's usability with tools that are not part of miden ecosystem itself, (things like cargo for instance, like in https://github.com/0xMiden/midenup/pull/118).
> Edit: The PR in question, editing CliCommand serialization is here:https://github.com/0xMiden/midenup/pull/133

Sounds good, no worries. I think we can utilize our internal Slack channel for better communication. Also, I missed that you self-assigned this issue, so that one is on me.

bitwalker:
Sorry it has taken me so long to follow up here, I've been pretty swamped as of late 😅.

> Do you mean doing something like ... If so, I think this would also work.

Yes, exactly. Since the faucet already depends on the client, it would seem natural then to have the `mint` command live there, and have it handle the relationship between the two crates.

> I don't think there would be a need to config the client instance necessarily (not in a way that it wasn't necessary before if you wanted to get assets into an account managed by yourself).

Agreed, and in particular, the client would have already been configured via `midenup` presumably (and/or by a prior use of the client via `miden`), and the faucet can simply piggy-back off of that configuration. I can't recall off the top of my head if the client has been set up to auto-initialize its state/configuration, or prompt the user to do so, but I'd expect that if the faucet calls into the client, then it could do the same there as well. If that initialization step is still being handled by `midenup`, then as long as the faucet component depends on the client component, then initialization of the client is guaranteed to happen before the faucet gets used anyway.

> Yes, I agree this is probably a pattern that would repeat in the future, and that splitting features out into a different tool could work out here

Fair. I think I'd like to wait and see another example of that pattern before we attempt to solve for it in `midenup`/`miden` though - not least because it will be clearer what the pattern actually is. For now, we only have this one scenario, that has a viable alternative implementation path, so it doesn't make for a particularly compelling use case IMO.

> Although in general, I'm not sure I fully understand how multi-binary command aliases could introduce new ways for the toolchain switching to break. As long as commands are their interfaces are tied to the correct versions, should it not be fine?

I'm more concerned about breaking changes to `midenup` itself and its manifest format, not the components themselves. By encoding more toolchain features/command handling in `midenup` (and necessarily then making its manifest more of a DSL), we significantly increase the likelihood of breaking and/or backwards-incompatible changes to `midenup` (and/or its manifest). My fear is that this would make toolchain upgrades and switching between older and newer toolchains take a dependency on specific version ranges of `midenup` itself, which defeats its purpose.

Consider `rustup`, which was the primary inspiration for `midenup` and its implementation - even with all the changes to Rust itself, and its toolchain components, I'm not aware of any requirement that you be running a specific version of `rustup` regardless of what version of Rust you want to work with (at least in the last several years). New `rustup` releases are very rare, because the actual meat of Rust is handled by the toolchain components, not `rustup`.

As I noted though, `miden` is part of `midenup` itself, and it may turn out that due to one reason or another, we'll need to split it out from `midenup` and treat it as its own component (likely one that is required/always installed). That comes with some downsides of its own, but I think that would be a better road to go down (if we _need_ to), than to compromise the stability/reliability of `midenup`'s toolchain management role.

lima-limon-inc:
> @lima-limon-inc I suspect this PR can be closed now, no?

Indeed, with https://github.com/0xMiden/midenup/pull/144 being merged; this can be closed.

Issue #131, opened 2025-12-09
feat: add `miden mint` ciommand


Keinberger:
Once this issue is implemented on the miden-faucet repository: https://github.com/0xMiden/miden-faucet/issues/192 , `midenup` should provide a `miden mint` alias command that performs the following 3 actions:

- execute `miden faucet mint`
- execute `miden client sync`
- execute `miden client consume-notes`

The result of this feature will be that users can mint fresh testnet tokens by only executing a single command.

In order for this command to work, `midenup` needs to be able to support running multiple commands as part of one alias, which is being tracked here: https://github.com/0xMiden/midenup/issues/129. Therefore, this issue here depends on the implementation of said issue.

PR #132, opened 2025-12-12
feat: implement multi-command aliases

### LINKED ISSUE #129: feat: aliases with multiple commands, from multiple binaries

Keinberger:
# Summary

This PR moves aliases to channel-level pipelines so a single alias can orchestrate multiple components. This was discussed in #129 . Each step now names both the `component` and the `command` tokens, and user arguments are appended only to the first step. `miden` resolves and executes every step in order, enabling cross-component flows (for example, `miden-faucet` followed by `miden-client`, i.e. [mint command](https://github.com/0xMiden/midenup/issues/131)).

Partial toolchains retain only those aliases which steps target installed components. The manifest (`manifest/channel-manifest.json`) has been updated to the new channel-level alias shape.

# Details

## Channel-level aliases (breaking change)

Aliases are no longer stored under individual components. Each channel now has a top-level `aliases` map (`HashMap<Alias, Vec<AliasStep>>`). Each `AliasStep` must specify a `component` (the component whose executable should run for this step) and a `command` (the existing `CliCommand` tokens: `executable`, `lib_path`, `var_path`, or verbatim strings). For example:

```json
"aliases": {
  "mint": [
    { "component": "faucet", "command": ["executable", "mint"] },
    { "component": "client", "command": ["executable", "sync"] },
    { "component": "client", "command": ["executable", "consume-notes"] }
  ]
}
```

Resolution prefers the active channel’s aliases and components and falls back to the installed channel with warnings. Partial installs keep aliases only if all steps target installed components.

## Multi-step execution

`miden` now resolves and runs each alias step sequentially. User CLI arguments are appended only to the first step, and a `--help` request is appended to the first step’s arguments. A helper resolves each step to `(program, args)` using the step’s `component`.

## Note on data passing

Data passing between steps (capture and templating) is not part of this PR, but required as a subsequent PR for implementing issue #131 to finally implement the `mint` command.

Closes #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ Here's a table with all the currently available aliases:
 | miden send       | Send transaction (state-changing) | miden-client send                                                    |
 | miden simulate   | Simulate transaction (no commit)  | miden-client exec                                                    |
 
+Aliases are defined at the channel level in `manifest/channel-manifest.json`. Each alias is a pipeline of steps, and each step must specify which component's executable to run. Commands use the existing `CliCommand` tokens (`executable`, `lib_path`, `var_path`, or verbatim strings).
+
+Pipeline steps execute in order. User-provided CLI arguments are appended to the first step only.
+
 
 ### Uninstalling a toolchain
 
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -23,7 +23,10 @@
           "name": "vm",
           "package": "miden-vm",
           "version": "0.15.0",
-          "features": ["executable", "concurrent"],
+          "features": [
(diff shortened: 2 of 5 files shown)


igamigo:
How does this implementation relate to the one of [#130](https://github.com/0xMiden/midenup/pull/130)?

PR #161, opened 2026-02-19
feat: introduce `alias_only` executables and check for `miden <library>`

### LINKED ISSUE #159: feat: restrict `cargo-miden` usage via `miden`
> I don't think we should allow invoking `cargo-miden` directly via `miden`, only subcommands e.g. `miden build` - that's not really changing here, but wanted to mention it.

_Originally posted by @bitwalker in https://github.com/0xMiden/midenup/issues/152#issuecomment-3874816683_

### LINKED ISSUE #160: bug: `miden` tries to execute `miden <library_name>` commands
Currently, when parsing `miden <name>` commands, we try to find a `Component` or `Alias` that shares the same name as `<name>`. However, we do not check if said component is a library. Therefore, after parsing, we try to execute `<name>`, which always fails. 

Therefore, if the name of a library is detected during parsing, `miden` should fail stating that `<name>` is not intended to be executed.

lima-limon-inc:
Closes #159 

`cargo miden` usage is now limited by the `alias_only` field. Now, if a component states `alias_only: true`, it will not be accessible via the `miden` interface.
Example:
```shell
$~: miden cargo-miden

Error: 'cargo-miden' is not intended to be called via 'miden', but rather by its aliases: 'new', 'build'

The Miden toolchain porcelain

Usage: miden <ALIAS|COMPONENT>

Available aliases:
  account
(...)
```

Closes #160

Now, if `miden <library name>` is entered in the CLI, `miden` will display a specific error message explaining why that operation is invalid ([here's](https://github.com/0xMiden/midenup/compare/fabrizioorsi/i158-limit-cargo?expand=1#diff-1a813ad5fe30a92311d801c75e869aa961e8e4f3605c553bee04c9dc4bc6c4b2R116-R119) the code).

```shell
$~: miden core

Error: 'core' installs the core.masp library. It is not intended to be executed as a binary.

The Miden toolchain porcelain

Usage: miden <ALIAS|COMPONENT>

Available aliases:
  account
  build
  call
  deploy
  faucet
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1152,8 +1152,14 @@ Error: {}",
 
         // Verify each executable component is accessible and runnable
         for component in &stable_channel.components {
+            let component_type = component.get_installed_file();
             // Skip libraries
-            if matches!(component.get_installed_file(), InstalledFile::Library { .. }) {
+            if matches!(component_type, InstalledFile::Library { .. }) {
+                continue;
+            }
+
+            // Skip components not meant to be executed directly
+            if matches!(component_type, InstalledFile::Executable { alias_only: true, .. }) {
                 continue;
             }
 


bitwalker:
In the future we could probably add a `miden run` command or something to that effect which executes things like `cargo-miden` that would otherwise only be accessible via an alias - but it's definitely not something we need now.

PR #192, opened 2026-04-17
fix: stores `Libraries` as `masl`

### LINKED ISSUE #191: bug: `Libraries` in `masp` format are saved as `masp`s.
When storing the `core` and `protocol` libraries, these were being stored as `masp`s instead of `masl`, as described by this comment: https://github.com/0xMiden/miden-vm/blob/eb5d43d82d61f6a18017a8c00e1a3695110b167d/crates/assembly-syntax/src/library/mod.rs#L404-L411.

As an aside, when trying to link these libraries with the compiler, I got the following error:
```shell
cargo miden build -L /Users/fabri/Library/Application\ Support/midenup/toolchains/0.14.3/lib -l core.masl

Creating Miden package miden_faucet_mint_tx.masp

failed to deserialize library from '/Users/fabri/Library/Application Support/midenup/toolchains/0.14.3/lib/core.masl': unexpected end of file
```

lima-limon-inc:
Closes #191 

Store libraries as `masl` instead of `masp`.

### DIFF
--- a/docs/src/getting-started/tutorial.md
+++ b/docs/src/getting-started/tutorial.md
@@ -34,5 +34,5 @@ A typical usage of midenup and miden might look like the following:
 
 7. With the generated Miden Package, an account can be created and deployed with the following command:
    ```shell title=">_ Terminal"
-   miden client new-account --account-type regular-account-updatable-code -p /path/to/package.masp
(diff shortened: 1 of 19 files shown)


bitwalker:
This is incorrect - "libraries" managed by midenup are intended to be packages, hence the `.masp` extension

PR #194, opened 2026-04-17
fix: migration bug

lima-limon-inc:
Currently, `midenup update` performed a migration when an upstream channel contained the `Migration` tag. However, the migration was triggered even when the channel had already been migrated.

This PR fixes this issue by only triggering a migration if the current channel's name matches the `Migration`'s `old_channel` field.

Additionally, the code was simplified a bit:
- `PartialEq for Channel` was removed since it was a bit too side-effect-y and was only used when finding a local `Channel`'s upstream counterpart. Similarly, `fn migrated_into` was also removed since.
- Instead, both of thses functionalities were encapsulated entirely inside the `find_upstream_counterpart` function.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 use crate::{
     artifact::{Artifacts, TargetTriple},
     config::Config,
+    manifest::Manifest,
     toolchain::{Toolchain, ToolchainJustification},
     utils,
     version::{Authority, GitTarget},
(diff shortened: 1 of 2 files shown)


bitwalker:
@lima-limon-inc Is this ready for final review/merge?

lima-limon-inc:
> @lima-limon-inc Is this ready for final review/merge?

It is ^_^. One small caveat though: there currently is an uninstall-related bug that I'm fixing over on https://github.com/0xMiden/midenup/pull/196 which causes `uninstall` to fail if a component was installed using an artifact.
This bug is also triggered by the migration pipeline.

lima-limon-inc:
> > @lima-limon-inc Is this ready for final review/merge?
> 
> It is ^_^. One small caveat though: there currently is an uninstall-related bug that I'm fixing over on #196 which causes `uninstall` to fail if a component was installed using an artifact. This bug is also triggered by the migration pipeline.

@bitwalker I've rebased this PR with the latests changes from main. It's now ready for review.

https://github.com/0xMiden/midenup/pull/201 is also ready for review and stems from this branch.

bitwalker (approved):

PR #201, opened 2026-05-07
refactor: atomic installation

### LINKED ISSUE #195: bug: `midenup` fails to uninstall component that was installed as a binary
Currently, when we try to uninstall a component, we delegate uninstallation to cargo entirely. 
However, if a component was installed as a binary artifact from the get go; cargo is not aware of its existence.

A potential fix would entail contemplating precompiled binaries inside the `uninstall_component` function

### LINKED ISSUE #198: proposal/refactor: make installs atomic
`midenup` saves the installed files directly into the user's filesystem. This means that, if an install gets stopped mid-way through, the filesystem is in an incomplete state. So, we take measures (such as logging) to figure out in which state the channel is in, in order to continue the install from there.

  However, this means that various parts of the codebase (namely, `uninstall.rs`) have to contemplate the possibility of an incomplete install, which adds complexity.

  This could be mitigated if installs where made atomic (meaning, the install is either fully installed, or not installed at all; no "incomplete" state). This could be achieved with the following changes:
  - Instead of installing the channel directly into the "final destination", it could be stored into a temporary "work directory" (that could either live in the OS's `tmpfs` of inside the `midenup` directory itself).
  - Once the channel is installed, we rename this "work directory" to the chanell's real name. If I'm not mistaken, the `rename` operation is atomic (at least in POSIX).
  - Afterwards, we update the manifest as usual. There is a bit of and edge case regarding the window of time between renaming the directory and updating the manifest, however this is something we already take into account.

  Codewise the changes are fairly minimal: the main change would happen in `install.rs` where we would create this work directory and execute the install into it. The install script itself would have basically no changes.

  With this, we could remove the `.installed_components` and `installation-successful` logic from `uninstall.rs` (as it would only be used in `install.rs`). Adittionally, we could also remove the `channel.json` file entirely and rely solely on the local manifest for uninstall.

lima-limon-inc:
Closes https://github.com/0xMiden/midenup/issues/198
Closes https://github.com/0xMiden/midenup/issues/195
Blocked-by: https://github.com/0xMiden/midenup/pull/209

This PR changes the way installs are performed in midenup. After this PR, installs in midenup are made atomically, meaning that the entire toolchain is either fully installed onto the filesystem or not.
         
Installs are performed on a separate directory that's stored inside the `installed_toolchains/<channel-version>-<hash>`. This directory is then **atomically** symlinked into the corresponding `toolchains/<channel-version>` directory. If a toolchain gets re-installed with newer components (such as in the case=
This structure roughly looks like:

```
midenup
├── installed_toolchains
│   └── 0.14.0-fb9b2252a909ddb9
│       ├── bin
│       │   └── (...)
│       ├── install.rs
│       ├── lib
│       │   └── (...)
│       ├── install.rs
│       └── opt
│           └── (...)
└── toolchains
    ├── 0.14.0 -> ../installed_toolchains/0.14.0-fb9b2252a909ddb9
    └── stable -> 0.14.0
```

Notes:
- A couple of trait implementations were removed in favor of the default derived ones. 
    - `Hash`, `PartialEq` and `Eq` for `Component`. These custom definition were only used in `update.rs` when performing set operations. Removed in 5597a1e53b70eee210300807c8c01896ffecac8e. 
    - Now, in order to compare `Component`s by names, the wrapper struct `ComponentByName` can be used.
- All components on the `Channel` now derive `Hash`, which is then used to generate the directory name in `installed_toolchains`.
- The `installation-successful` and `.installation-in-progress` install-logs have been removed. Since toolchains are atomically installed, if the toolchain is symlinked, then it was definitely installed.
- The `channel.json` snapshot is no longer saved during installs. Now, `uninstall` relies on the information saved on the local manifest to uninstall the component. This got changed on 2c39912e71ba11020c4234230afb78c36aaaf5d
    - This change is what fixes https://github.com/0xMiden/midenup/issues/195.
- There's one limitation with this approach: if a channel requires a name change when updating, the install is no longer "atomic".
    - This particularly affects channel migrations. However, as pointed out in [the code](https://github.com/0xMiden/midenup/pull/201/changes/f9967ef7683a93a4a842e1199a454bdb314214ee#diff-4bf1ef1a8771c506c71d2ef69d5d59e623f54b2cc09f2c297496e30d5f7da25eR203-R206), were a migration to be stopped _right before_ installing the previous channel, re-issuing a `midenup update` would finish the process accordingly.


lima-limon-inc:
@bitwalker like mentioned on https://github.com/0xMiden/midenup/pull/194, this PR is also ready for review. The main set of changes come from `install.rs` + `channel.rs`.

Thanks in advance ^_^

tests/data/integration_migration_test/channel-manifest-2.json
```diff
@@ -19,13 +19,6 @@
           "installed_library": "core.masp",
           "library_struct": "miden_core_lib::CoreLibrary"
         },
-        {
-          "name": "vm",
-          "package": "miden-vm",
-          "version": "0.20.3",
-          "features": ["executable", "concurrent"],
-          "installed_executable": "miden-vm"
-        },
```
lima-limon-inc, line 28: This wasn't removed from anything in particular other than to save time for the tests.

src/commands/install.rs
```diff
@@ -114,16 +133,43 @@ pub fn install(
...
+    // tmp_link is a symlink file that points to relative_install_target. Even
+    // if tmp_link file is moved, it will still point to relative_install_target.
+    // For further reference on atomic directory updates, see:
+    // https://axialcorps.wordpress.com/2013/07/03/atomically-replacing-files-and-directories/
+    utils::fs::symlink(&temp_symlink, &relative_install_target)?;
+
+    // We now rename tmp_link to toolchain_dir. When renamed, it will still be
+    // pointing to relative_install_target. If the channel direcotry existed, it
+    // will overwrite the file. This is what marks the install as completed.
+    std::fs::rename(&temp_symlink, &toolchain_dir).with_context(|| {
```
bitwalker, line 157: It isn't clear to me whether there will be an issue here with existing installations. Those may already have real directories at `toolchains<version>`, so this will fail if `toolchain_dir` already exists.

src/commands/uninstall.rs
```diff
@@ -17,164 +17,144 @@ use crate::{
...
+            // uninstallation attempt that got interrumpted.
+            && stable_symlink.exists()
+        {
             std::fs::remove_file(stable_symlink).context("Couldn't remove symlink")?;
         }
     }
 
-    local_manifest.remove_channel(channel.name.clone());
+    // We remove the symlink, thus making the channel unaccesible.
+    if toolchain_symlink.exists() {
```
bitwalker, line 90: Similar to my comment in `install.rs`, if there is an existing install, then `remove_file` here may be getting called on a real directory, not a symlink. Existing installs will need a one-time migration or compatibility path.

src/commands/install.rs
```diff
@@ -22,39 +27,53 @@ pub fn install(
...
+    let toolchain_dir = toolchains_dir.join(format!("{}", &channel.name));
 
-    let installation_indicator = toolchain_dir.join("installation-successful");
-    let is_partial = local_manifest
-        .get_channel_by_name(&channel.name)
-        .map(|ch| ch.is_partially_installed())
-        .unwrap_or(false);
+    let installed_toolchains_dir = config.midenup_home.join("installed_toolchains");
+    let install_dir_name = format!("{}-{}", &channel.name, channel.content_hash());
+    let install_dir = installed_toolchains_dir.join(&install_dir_name);
```
bitwalker, line 36: It appears updates that have the same hash can be marked complete without reinstalling. Same-hash updates can be marked complete without reinstalling. `install_dir` is based on `channel.content_hash()` here, and changed branch/path components can reuse the same upstream channel hash. If the hashed install directory already exists, component removal is skipped entirely (see line 44), the generated install script skips existing binaries, and the manifest is refreshed as if the update happened. This affects branch-tip and local-path updates.

src/commands/update.rs
```diff
@@ -152,125 +155,70 @@ fn update_channel(
...
-
-                    *component_to_install = local_component.clone();
-                }
+        let do_update = match component.get_installed_file() {
+            InstalledFile::Library { .. } => true,
+            InstalledFile::Executable { .. } => match component.version {
+                Authority::Cargo { .. } | Authority::Git { .. } => true,
+                // Since uninstalling a component from the filesystem is potentially
+                // irreversible, we take special precautions before uninstalling them.
+                Authority::Path { .. } => match options.path_update {
```
bitwalker: `--path-update=off` no longer preserves skipped components. Update excludes path components here, but installation still saves the upstream channel and refreshes path metadata for all path components (see line 223 of `install.rs`), so a skipped rebuild can then look current.

src/commands/install.rs
bitwalker, line 157: It looks like installs can be published before metadata is committed. The runnable `toolchains/<version>` symlink is published before `stable` and `manifest.json` are written here and on line 176. A failure in between will leave the filesystem and local manifest state inconsistent.

src/commands/uninstall.rs
bitwalker, line 90: It appears uninstallation is not retry-safe after cleanup failure. This removes the symlink and manifest entry before removing components/install directory. If cleanup fails, rerunning uninstall bails because the manifest entry is gone.

bitwalker (changes requested):
Looks good overall! I found a couple of issues that we should address, but this is pretty close

src/commands/install.rs
lima-limon-inc: Great point, addressed it here: https://github.com/0xmiden/midenup/compare/35b49ac..970b91d

I added a migration check and associated function. Were more migrations to be needed in the future, we could just add more checks and associated fuctions on the corresponding `mod.rs` file

src/commands/install.rs
lima-limon-inc: Great catch, I've since adressed this issue here: https://github.com/0xmiden/midenup/compare/2402558..725ac8f

Now, when performing an update, the channels *might* mutate (for instance, channels that are tracking `branch`es will get the fetched latest commit added onto their `latest_commit` field. Same with `Path` tracked components.

src/commands/update.rs
lima-limon-inc: Noted. This has seen been fixed here: https://github.com/0xMiden/midenup/pull/201/commits/f29ac1eca5544c89f7e6e51e925fe97486b532e4

This same issue also caused a bug where partial channels would install all the remaining components.

src/commands/install.rs
lima-limon-inc: This has since been fixed here: https://github.com/0xMiden/midenup/pull/201/commits/d58c9dcaf4fc23d3a10d3da006b5546e8223b33d

Updating the manifest should be the last operation we do IIUC

src/commands/uninstall.rs
lima-limon-inc: Good point, changed here: https://github.com/0xMiden/midenup/pull/201/commits/f1a3b5dff88878b4a7de6a6c298538515eca049f

Now the symlink is removed after removing the components.

src/commands/install.rs
lima-limon-inc: Update: This is no longer the case, the `is_up_to_date` function no longer changes the component. Now, when the `UpstreamChannel` is created, we sync the channel to get the latest changes.

lima-limon-inc:
The latest commits updated the way updates were computed. The main change is the following:
https://github.com/0xmiden/midenup/compare/fd96e357560fc4a3f33b94db1de4df69c2dee28b..f1a3b5dff88878b4a7de6a6c298538515eca049f#diff-4bf1ef1a8771c506c71d2ef69d5d59e623f54b2cc09f2c297496e30d5f7da25eR487-R508
- The old channel gets cloned as "base", in order to inherit tags
- The components are taken from the `components_to_install` variable, which includes both the components that need updating and, notably, the components that are up to date
- If the channel got migrated, we apply the changes (i.e. the name change), on the `migrate_channel` function


The entire change set can be seen on this diff: https://github.com/0xmiden/midenup/compare/fd96e357560fc4a3f33b94db1de4df69c2dee28b..aab07d5c0c32eb76c113d61f39960f943dacc840

Issue #171, opened 2026-03-19
Installation friction: PATH, zshrc, progress, toolchain warning


Keinberger:
## Context

Multiple issues reported during Miden Day related to midenup installation experience.

## Items

- [x] PATH directives in docs are wrong. Should point to `toolchains/stable/bin`, not `midenup/bin`.
- [ ] `~/.zshrc` doesn't exist on fresh macOS, midenup setup fails silently.
- [ ] `midenup install stable` takes 20-30 minutes with no progress indicator. Users ctrl-C thinking it's stuck.
- [x] "WARNING: new is an alias from component cargo-miden, which is installed but is not part of the current active toolchain" is confusing for both humans and AI agents.

lima-limon-inc:
Update on this issue:

- PATH directives in docs are wrong. Should point to `toolchains/stable/bin`, not `midenup/bin`: This has been addressed in https://github.com/0xMiden/docs/issues/181
- [ ] `~/.zshrc` doesn't exist on fresh macOS, midenup setup fails silently.
- [x] `midenup install stable` takes 20-30 minutes with no progress indicator. Users ctrl-C thinking it's stuck. This is being tracked here: #177 
- [x] "WARNING: new is an alias from component cargo-miden, which is installed but is not part of the current active toolchain" is confusing for both humans and AI agents. This bug was addressed here: https://github.com/0xMiden/midenup/pull/173. 

For the addressed items, I'm marking them as done.

bitwalker:
I think we should show the output from `cargo` when installing components via `cargo install` - it will always be an option to install components that way, and we need useful output. There's not really a useful way to measure progress though, so it's best to simply show what Cargo is doing IMO.

PR #202, opened 2026-05-12
docs(init): fix PATH guidance for midenup install

Jr-kenny:
## Summary
- update the install docs to point users at the `miden` symlink created in `$CARGO_HOME/bin` instead of the stale `$MIDENUP_HOME/bin` path
- clarify the zsh/macOS flow by explicitly mentioning `~/.zprofile` creation
- update `midenup init`'s PATH warning text to match the actual install layout

## Why
`midenup init` creates a `miden` symlink under Cargo's bin directory, but the docs and warning text still send users toward paths that do not match the current setup. That makes first-run installation more confusing than it needs to be.

## Testing
- `git diff --check`
- did not run Rust tests locally because this machine does not have `cargo` installed

Related to #171

### DIFF
--- a/docs/src/getting-started/installation.md
+++ b/docs/src/getting-started/installation.md
@@ -54,7 +54,15 @@ source ~/.zprofile
 
 ### For Bash
 
-Add the following to your `~/.bash_profile` file:
+Add the following to your `~/.bash_profile` file (use `~/.bashrc` on Linux). If the file does
+not exist yet, create it first:
+
+```bash title=">_ Terminal"
+touch ~/.bash_profile
+echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> ~/.bash_profile
+```
+
+If you prefer to edit the file manually, add:
 
 ```bash title=">_ Terminal"
 export PATH="$HOME/.cargo/bin:$PATH"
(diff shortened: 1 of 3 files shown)


docs/src/getting-started/installation.md
```diff
@@ -18,27 +18,32 @@ Until this crate has been published to crates.io, it is only possible to install
...
 
 ### For Zsh (macOS default)
 
-Add the following to your `~/.zprofile` file:
+Add the following to your `~/.zprofile` file. If the file does not exist yet, create it first:
 
 ```bash title=">_ Terminal"
-export MIDENUP_HOME="/Users/$(whoami)/Library/Application Support/midenup"
-export PATH=${MIDENUP_HOME}/bin:$PATH
+touch ~/.zprofile
```
bitwalker, line 39: This presumes a specific shell - if we're going to do this, we should provide instructions for both `bash` and `zsh`.

bitwalker (changes requested):
Thanks, one tweak and this can be merged

docs/src/getting-started/installation.md
Jr-kenny: good call, added the same touch+echo helper for bash with a note about ~/.bashrc on linux. pushed in 5460c70

bitwalker (approved):

bitwalker:
Note for the future - we require signed commits for contributions. I'll merge this anyway this time since this is your first contribution and it's just a doc change, but next time please sign them.

PR #213, opened 2026-06-23
chore: add v0.15 to channel manifest

TomasArrachea:
Closes https://github.com/0xMiden/midenup/issues/212

Updates the `channel-manifest.json` to add the new v0.15 release artifacts. Notably, the miden-node binary has now been divided into 4 separate binaries (`node`, `validator`, `ntx-builder`, `remote-prover`).

This manifest can be tested by running: 
```
MIDENUP_MANIFEST_URI=file://<path-to-new-manifest> midenup install stable
```

TODO:
- ~Re-add the `miden start-node` alias. We need to define where should the node startup logic live, since the `bundled` command does not exist anymore~. EDIT: This has been decided to be removed from midenup, docker-compose will fill this use.
- Decide if we want to keep all the node binaries. Keeping them would increase installing times and might not be useful for the average user.

### DIFF
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -665,19 +665,7 @@
           "name": "node",
           "package": "miden-node",
           "version": "0.15.0",
-          "installed_executable": "miden-node",
-          "aliases": {
-            "start-node": [
-              "executable",
-              "bundled",
-              "start",
-              "--data-directory",
-              "var_path",
-              "data",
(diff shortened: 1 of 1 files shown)


bitwalker:
I've manually applied the v0.15 toolchain due to the conflicts with `main`, as well as adding in a few changes (e.g. artifacts for some of the components that now have them, and using the latest versions of some components that have had patch releases since this was opened).

PR #219, opened 2026-08-01
feat: make 'miden debug' project-aware

djolertrk:
Automatically discovers the built `.masp` package and `inputs.toml`, building the project first when needed. Explicit debugger inputs remain unchanged.


bitwalker (changes requested):
I don't think `miden` should be doing things like this - it binds `midenup`/`miden` to specific semantics associated with a given toolchain, which is explicitly something it is _not_ supposed to do.

This logic belongs in `miden-debug` IMO - where we can already trivially detect if we're being run under `miden`/`midenup`, and can be much more precise about how we determine what to inject (and when).

I would like to keep the patch that respects an explicit `MIDENUP_HOME` override though - that's definitely a gap in the current code, and is something that `miden`/`midenup` necessarily need to be the arbiter of.

PR #220, opened 2026-08-01
midenup: manifest v3, support network aliases

bitwalker:
This PR is an alternative approach to #218 that solves a handful of related issues, and also bumps the manifest version to 3.0 in order to ensure older clients are forced to upgrade now that 1.0.0-alpha.1 is released.

Primary differences:

* The networks map is a top-level item, not per-channel. It maps a network name to the channel it currently aliases
* The `update-manifest` tool now supports promoting networks to new toolchains
* Various bugs with manifest v2 implementation were solved (reading the manifest cache on every command, paths being able to reach outside of `MIDENUP_HOME`, and others)

The one thing not implemented here that we need to build is to have `midenup show` when a local network link is behind the upstream manifest, so that a user knows to upgrade. I'd have you start on that next @TomasArrachea.

I ended up going through all the design and implementation on this sort of by accident this week in the course of trying to work out how best to do it, but am happier overall with where this ended up compared to #218 in its current form. I'd suggest we review this and plan to merge it + release 1.0.0-beta.1 early next week, and let it supercede #218 - but we can discuss tradeoffs if you feel there is a good argument to be made for the approach taken there.


docs/src/design/manifest.md
```diff
@@ -43,19 +43,41 @@ A channel is a set of [components](#components) under one version, meant to be u
...
+  "devnet":  "0.16.0",
+  "mainnet": "0.15.0",
+  "testnet": "0.16.0"
+}
+```
+
+Several networks may name one channel, which is the normal state once a testnet toolchain is
+promoted to mainnet.
+
+`stable`, `beta` and `nightly` are accepted as synonyms for `mainnet`, `testnet` and `devnet`. They
```
TomasArrachea, line 70: nit: I'm not sure if having these synonyms helps for clarity. I would remove them, but just a preference.

src/commands/update.rs
```diff
@@ -94,11 +52,106 @@ pub fn update(
...
+        IntentUpdate::Replace(installation.intent.clone()),
+        Changes::default(),
+        options,
+        // This command owns the pointer move, so DERIVE must not make it: see below.
+        Some(name),
+    )?;
+
+    // The user's data follows the network it belongs to, rather than being stranded under a channel
+    // version they are no longer tracking. A rename, so it cannot half-happen.
+    let carried = carry_var_to(&config.midenup_home, &installed, &target)?;
```
TomasArrachea: This moves `var/<old>` to `var/<new>` whenever the destination is free, but `var/` is keyed by channel and several networks may name one channel, so I think that updating one network would move the shared directory out from under the others.

src/commands/install.rs
```diff
@@ -153,6 +174,42 @@ pub fn install(
...
 }
 
+/// The warning owed to a user whose network is about to name a different channel while their data
+/// is still under the one it named before.
+///
+/// **Deliberately a warning rather than a carry.** Carrying would make an install mutate data it
+/// did not create, and `install` has no notion of "the channel you were on" -- only `update
+/// <network>` does, which is why that is where the carry lives. Before networks existed the two
+/// commands agreed by both doing nothing; now they perform the same pointer move, so the difference
+/// has to be said out loud rather than left silent.
```
TomasArrachea: I'd avoid references to prior implementations in the comments in general.

src/commands/install.rs
```diff
@@ -153,6 +174,42 @@ pub fn install(
...
+    }
+    if !paths::var_dir(home, &previous).is_dir() {
+        return None;
+    }
+
+    Some(format!(
+        "{}: {network} now names {channel}; your data for {previous} is still at \
+         var/{previous}.\n         Run 'midenup update {network}' to carry it across.",
+        "warning".yellow().bold(),
+    ))
```
TomasArrachea: This tells the user to run `midenup update <network>` to carry the data, but since the "derive" step will repoint the link, the update command will early return and not carry the data (this is described in the `InstallationOptions::reconciling` doc comments). We should either fix this or change the warning message.

docs/src/design/manifest.md
bitwalker: I largely agree with you there. I will note that they are basically invisible synonyms (i.e. the actual toolchain aliases used on disk are the network names, and these names never show up in the CLI). Their purpose for remaining around was as a way for people with muscle memory for those names to get more or less what they are expecting to get toolchain-wise (i.e. `beta` mapping to `testnet` and `nightly` mapping to `devnet` are essentially correct characterizations of what those networks are used for, staging and development respectively).

Maybe we keep them around for the initial release to make migration easier, and then remove them before we publish the final `1.0.0` release?

src/commands/update.rs
bitwalker: Good catch!

src/commands/install.rs
bitwalker: I've done a broader sweep to clean up comments of this sort and found a few inaccuracies as well, so I appreciate the note!

src/commands/update.rs
bitwalker: I've revisited how `var` is keyed in general, since it occurred to me that we actually want a slightly different property from `var` than other parts of the toolchain. Namely, when it comes to networks, we intuitively would expect to have isolated storage between them, but `midenup` was sharing storage between networks when they were pinned to the same channel. I've reworked this so that `var` is now keyed to the channel the user named (so if the user named `mainnet`, then `var/mainnet` reflects the storage associated with that network, even across channel upgrades; and a versioned channel like `0.15.0` goes to `var/0.15.0` as you'd expect, but is never shared with a network that happens to be pinned to the `0.15.0` toolchain).

src/commands/install.rs
bitwalker: This is made moot now by the changes I made as a result of your other comment (see my note there for more details), and this code and related machinery is removed. Two things to note:

* `carry_var_to` sticks around for channel migration only
* `--purge` now keys on the user-provided selector, so uninstalling a channel can never remove a network's store (because they aren't the same)

docs/src/design/manifest.md
TomasArrachea: Sounds good.

src/paths.rs
```diff
@@ -49,21 +52,41 @@ pub fn toolchain_link(home: &Path, channel: &semver::Version) -> PathBuf {
...
+    toolchains_dir(home).join(network)
+}
+
 /// Where installed trees live.
 pub fn publications_dir(home: &Path) -> PathBuf {
     home.join("publications")
 }
 
-/// Mutable, component-owned state for a channel: `%var`.
+/// Mutable, component-owned state: `%var`. The Miden client's database lives here (`%var(data)`).
```
TomasArrachea, line 69: Currently `var/` would not actually hold the client database (in the manifest the client aliases don't reference `%var`).  Miden client resolves to a global `.miden` directory, so we could point that at the `var/` directory by setting the `MIDEN_CLIENT_HOME` env.

But, do we actually want the `var/` directory to hold the client database? The client already has mechanisms for managing that directory, and it knows which network it connected to. Delegating the data management to it could be an option. This way midenup only would track the toolchain being used, and not the client state.

bitwalker: That's actually preferable IMO (the client doing its own thing) - the purpose of `var` is more about toolchain-specific volatile data storage anyway, and if the client has its own way of managing that separately, then all the better, since that takes the burden off of `midenup` to "do the right thing". In general I would expect `var` to be used by toolchain components that require per-toolchain isolation, but aren't necessarily "toolchain-aware". The client was really the most obvious example of this, but since it handles that complexity itself, that keeps things simpler.

I'll update the documentation though to make it clear that the client doesn't actually store its database in `var` today, but that it serves as an example of why the implementation works the way it does.

docs/src/design/midenup.md
```diff
@@ -18,19 +18,25 @@ $MIDENUP_HOME/
...
+**`var/` is outside the publication, and keyed by the selector.** It holds mutable component-owned state — most importantly the Miden client's database, referenced from the manifest as `%var(data)`. Its key is what the *user selected* — `var/mainnet`, `var/0.15.0` — not the channel that selector resolves to. Networks are logically distinct even when several of them name one toolchain, which is the usual state of affairs: your mainnet accounts and your testnet notes are not one database. A selector also does not move when a pointer moves, so a network's store stays exactly where it is as the network advances. Install, update, republication and pointer moves never read, write, move or delete it. The one exception is channel migration, which renames `var/<old>` to `var/<new>` — both pinned-version selectors, and migration is what retires one.
```
TomasArrachea: We should also update this reference.

TomasArrachea (approved):
LGTM!

PR #222, opened 2026-08-04
feat: report when local network link drifts from upstream

TomasArrachea:
Closes https://github.com/0xMiden/midenup/issues/221.

### DIFF
--- a/docs/src/design/spec.md
+++ b/docs/src/design/spec.md
@@ -779,7 +779,7 @@ The comparison is **inequality, not "is newer"**. The pointer is authoritative i
 
 Repointing the link is done by this command rather than left to the DERIVE step of the install it performs, because an update whose target is already installed can legitimately have nothing to install - and moving the pointer is the thing this command exists to do.
 
-A link that disagrees with upstream is reported by `midenup show list` on the channel it names, with this command as the remedy - `0.14.0 (mainnet is now 0.15.0 upstream - run midenup update mainnet)` - so a user learns that a network has moved.
+A link that disagrees with upstream is reported by `midenup show list` on the channel it names, with this command as the remedy - `0.14.0 (mainnet is now 0.15.0 upstream -- run 'midenup update mainnet')` - so a user learns that a network has moved.
 
 ---
 
--- a/docs/src/design/spec.md
+++ b/docs/src/design/spec.md
@@ -779,6 +779,8 @@ The comparison is **inequality, not "is newer"**. The pointer is authoritative i
 
 Repointing the link is done by this command rather than left to the DERIVE step of the install it performs, because an update whose target is already installed can legitimately have nothing to install - and moving the pointer is the thing this command exists to do.
 
+A link that disagrees with upstream is reported by `midenup show list` on the channel it names, with this command as the remedy - `0.14.0 (mainnet is now 0.15.0 upstream - run midenup update mainnet)` - so a user learns that a network has moved.
+
 ---
 
 ## 12. Migration from manifest v1.0.1
--- a/docs/src/getting-started/usage.md
+++ b/docs/src/getting-started/usage.md
@@ -42,7 +42,7 @@ To list all the currently installed toolchains in the system, run:
 midenup show list
 ```
 
-Each toolchain is listed with the networks that run it, and a network that has since moved to a different toolchain is listed with the command that catches you up: `0.14.0 (mainnet is now 0.15.0 upstream -- run midenup update mainnet)`.
+Each toolchain is listed with the networks that run it, and a network that has since moved to a different toolchain is listed with the command that catches you up: `0.14.0 (mainnet is now 0.15.0 upstream -- run 'midenup update mainnet')`.
 
 ## Using a toolchain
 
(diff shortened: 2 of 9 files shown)


docs/src/design/spec.md
```diff
@@ -779,6 +779,8 @@ The comparison is **inequality, not "is newer"**. The pointer is authoritative i
 
 Repointing the link is done by this command rather than left to the DERIVE step of the install it performs, because an update whose target is already installed can legitimately have nothing to install - and moving the pointer is the thing this command exists to do.
 
+A link that disagrees with upstream is reported by `midenup show list` on the channel it names, with this command as the remedy - `0.14.0 (mainnet is now 0.15.0 upstream - run midenup update mainnet)` - so a user learns that a network has moved.
```
bitwalker:
```suggestion
A link that disagrees with upstream is reported by `midenup show list` on the channel it names, with this command as the remedy - `0.14.0 (mainnet is now 0.15.0 upstream -- run 'midenup update mainnet')` - so a user learns that a network has moved.
```

docs/src/getting-started/usage.md
```diff
@@ -42,6 +42,8 @@ To list all the currently installed toolchains in the system, run:
 midenup show list
 ```
 
+Each toolchain is listed with the networks that run it, and a network that has since moved to a different toolchain is listed with the command that catches you up: `0.14.0 (mainnet is now 0.15.0 upstream -- run midenup update mainnet)`.
```
bitwalker:
```suggestion
Each toolchain is listed with the networks that run it, and a network that has since moved to a different toolchain is listed with the command that catches you up: `0.14.0 (mainnet is now 0.15.0 upstream -- run 'midenup update mainnet')`.
```

src/networks.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::{collections::BTreeMap, path::Path};
+
+use crate::paths;
+
+/// Every `toolchains/<network>` link on this machine, with the channel it names.
+///
+/// The inverse of [`paths::network_link`]. An unreadable or absent `toolchains/` is an empty map
+/// rather than an error: every caller is reporting, and has nothing to say about a home with no
+/// links in it.
+pub fn links(home: &Path) -> BTreeMap<String, semver::Version> {
```
bitwalker, line 14: The scanning logic is now duplicated. `uninstall.rs` (line 66) has its own `toolchains/` walk with different rules — no `default` skip, exact bare-version compare, no semver parse. There's now two answers to "which entries are network links", and `paths.rs`'s module docs warn about this ("spelling a layout path inline is how install and uninstall drifted apart"). `uninstall` should use `networks::links()` too.

src/networks.rs
```diff
@@ -0,0 +1,169 @@
...
+        let Ok(target) = std::fs::read_link(entry.path()) else {
+            continue;
+        };
+
+        // The whole target is parsed as a version, which is what makes this rule sufficient on
+        // its own: a version string cannot contain a separator, so this admits only a
+        // single-segment relative target. Everything else in the directory is excluded by
+        // it -- a channel's own link names `../publications/<channel>-<id>`, a tombstone
+        // names `.uninstalled`, and `default` is absolute under either override form.
+        let Some(channel) = target.to_str().and_then(|target| semver::Version::parse(target).ok())
```
bitwalker: Drift reporting silently doesn't apply when the network link is absolute.

`networks::links()` parses the whole target; every other reader of these same links uses `.file_name()`: `config.rs:240` (`local_channel`), `config.rs:196`, `migrate_networks.rs:90/190`. So on a home with an absolute link:

```
0.14.0                            <- no annotation, no drift report
local_channel resolves it anyway: Some(0.14.0)
```

Reachable via a pre-alpha home, where `install` wrote `toolchains/stable` absolutely; `migrate_networks` renames it preserving the target, yielding an absolute `mainnet`. Everything else keeps working, so the feature just quietly doesn't work for those users.

The fix isn't simply `file_name()`, `toolchains/0.15.0 -> ../publications/0.15.0-abc123` has file name `0.15.0-abc123`, which parses as a valid semver with prerelease `abc123`, so a filename fallback would read a channel's own link as a network link naming a channel that doesn't exist. It needs the filename parse plus a check that the target is a single segment or resolves directly inside `toolchains/`.

The note on the line above this, i.e. "default is absolute under either override form" is inaccurate:  `migrate_networks::repoint_default` writes a relative `default -> mainnet` when the link it replaced was relative. It's harmless because the name check on line 26 runs first, but might be worth fixing as well.

bitwalker (approved):
This looks good! I left a couple of nits that we should clean up before this is merged, but nothing big. Nice work!

src/networks.rs
TomasArrachea: Fixed in [d892dd6](https://github.com/0xMiden/midenup/pull/222/commits/d892dd6437d96cba99354fb0e13c9888be115f27)!

src/networks.rs
TomasArrachea: Updated `uninstall.rs` to use`networks::links()` as well.

PR #226, opened 2026-08-11
feat: support tar.gz artifacts

### LINKED ISSUE #225: Support compressed artifacts
Currently `midenup` fetches artifacts and installs them directly. We should be able to interpret compressed artifacts, so midenup decompresses them on installation.

TomasArrachea:
Closes https://github.com/0xMiden/midenup/issues/225.

This PR:
- Introduces support for artifacts that are packaged inside a `tar.gz` file. 
- Updates the `manifest-channel.json` to re-add `cargo-miden` and `midenc` to the v0.16.0 channel since they are now supported.

### DIFF
--- a/src/install/archive.rs
+++ b/src/install/archive.rs
@@ -63,7 +63,7 @@ fn extract_bounded(
     // One past the limit, so that running out of allowance is distinguishable from a stream that
     // ended on it.
     let mut stream = match format {
-        SupportedFormat::TarGz => flate2::read::GzDecoder::new(body).take(limit + 1),
+        SupportedFormat::TarGz => flate2::read::MultiGzDecoder::new(body).take(limit + 1),
     };
 
     let result = from_tar(&mut stream, uri, out);
@@ -91,6 +91,10 @@ fn from_tar<R: Read>(stream: R, uri: &str, out: &mut impl Write) -> Result<(), A
     };
 
     let mut tar = tar::Archive::new(stream);
+    // A tar end marker is allowed to be followed by more tar data. Continue past zero blocks so a
+    // concatenated archive cannot hide a second file, and so arbitrary trailing payload is parsed
+    // and rejected rather than silently discarded below.
+    tar.set_ignore_zeros(true);
     let entries = tar.entries().map_err(malformed)?;
 
     // A flag rather than the bytes: the file is gone once written out, and a decompressing stream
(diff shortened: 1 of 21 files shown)


src/install/archive.rs
```diff
@@ -0,0 +1,165 @@
...
+    let entries = tar.entries().map_err(|err| ArchiveError::Malformed {
+        uri: uri.to_string(),
+        reason: err.to_string(),
+    })?;
+
+    // Held as bytes rather than as a position, because a decompressing stream cannot be rewound to
+    // come back for it.
+    let mut only: Option<Vec<u8>> = None;
+
+    for entry in entries {
```
bitwalker, line 106: Please validate the gzip stream through EOF before publishing. `tar::Archive::entries()` stops at the tar end blocks, so the `GzDecoder` is dropped without checking the gzip CRC/ISIZE footer. Removing the final 8 bytes from a real release archive makes `gzip -t` reject it, but this code still extracts and installs it successfully. Drain the decoder after scanning the tar entries and reject footer/checksum/trailing-stream errors before returning the artifact.

src/install/archive.rs
```diff
@@ -0,0 +1,165 @@
...
+        if !entry.header().entry_type().is_file() {
+            continue;
+        }
+
+        // A second file settles it, so there is nothing to learn from the rest of the stream.
+        if only.is_some() {
+            return Err(ArchiveError::NonZeroFiles { uri: uri.to_string() });
+        }
+
+        let mut body = Vec::new();
```
bitwalker: Please avoid buffering the entire decompressed member in an unbounded `Vec`. The downloader still owns the complete compressed body while `read_to_end` allocates the complete extracted file. The Linux `cargo-miden` asset added by this PR is 69,319,367 bytes compressed and 327,144,760 bytes unpacked, so a normal install needs at least ~396 MB simultaneously before allocation overhead; a gzip bomb is unbounded. Stream extraction to the temporary destination and enforce a decompressed-size limit.

src/manifest/validate.rs
```diff
@@ -294,14 +316,67 @@ fn validate_names(channel: &Channel, errors: &mut Vec<ValidationError>) {
...
+    artifact: &crate::artifact::Artifact,
+    component: &Component,
+) -> Option<&'static str> {
+    // Best effort: an artifact whose URI cannot be resolved at all has its own errors, and platform
+    // -specific resolution is not this validator's business.
+    let uris = artifact.get_uris_for(id, component).ok()?;
+
+    crate::artifact::ArchiveFormat::supported_spellings().find(|spelling| {
+        let suffix = format!(".{spelling}");
+        !id.ends_with(&suffix) && uris.iter().any(|uri| uri.to_string().ends_with(&suffix))
```
bitwalker: Please inspect the URI path rather than applying `ends_with` to the complete URI string. A valid source such as `https://host/tool.tar.gz?download=1` (or one with a fragment) passes this check without an `archive` declaration, after which midenup installs the container as the artifact. Parse the HTTP URL path while continuing to treat `file://` sources as filesystem paths.

bitwalker (changes requested):
This looks mostly good, but a handful of issues were found with the new archival code that need to be fixed before we can merge. Technically older manifest v3 clients will not properly handle this field (i.e. they will download and install the tarball _as_ an executable), and will not be notified that they need to update - but since the releases so far have been beta releases, I think we can just focus on getting this merged, doing a new release, and redirecting anyone that reports an issue due to installing the beta to upgrade

src/install/archive.rs
TomasArrachea: Done. Now the buffer is fixed at 64 KiB.

bitwalker (approved):

PR #232, opened 2026-08-12
docs: add PowerShell PATH setup steps

### LINKED ISSUE #231: Add PowerShell PATH instructions to the installation guide
The installation guide has detailed PATH setup steps for zsh and bash, but the PowerShell section is still a placeholder:

```md
:::note todo
Add instructions here
:::
```

This leaves Windows users without the equivalent command for adding Cargo's bin directory to their user PATH after running `midenup init`.

A small docs fix could replace the TODO with PowerShell instructions for adding `$HOME\.cargo\bin` to the user PATH and reloading the shell.

omerbek:
Summary
Adds the missing PowerShell instructions for putting Cargo's bin directory on the user PATH during midenup installation.

Fixes #231.

Why
The installation guide already has zsh and bash PATH setup steps, but the Windows PowerShell section was still a TODO. That leaves Windows users without the equivalent setup path after `midenup init` creates the `miden` command in Cargo's bin directory.

Changes
- Replaced the PowerShell TODO with commands that add `$HOME\.cargo\bin` to the user PATH if it is not already present.
- Added a short verification step using `miden help`.

Testing
- `git diff --check`
- `npm run build:dev` in `docs/`

### DIFF
--- a/docs/src/getting-started/installation.md
+++ b/docs/src/getting-started/installation.md
@@ -80,9 +80,25 @@ If you forget to do the step above, some functionality will not work as expected
 
 ### For PowerShell (Windows)
 
-:::note todo
-Add instructions here
-:::
+Add Cargo's bin directory to your user `Path` environment variable:
+
+```powershell title=">_ Terminal"
+$cargoBin = Join-Path $HOME ".cargo\bin"
+$userPath = [Environment]::GetEnvironmentVariable("Path", "User")
+$paths = @($userPath -split ";" | Where-Object { $_ })
+
+if ($paths -notcontains $cargoBin) {
+    $newPath = ($paths + $cargoBin) -join ";"
+    [Environment]::SetEnvironmentVariable("Path", $newPath, "User")
+}
+```
+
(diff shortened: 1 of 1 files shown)


bitwalker (changes requested):
I appreciate the PR, but I'm hesitant to add any documentation that suggests we support WIndows at the moment:

1. We have no CI coverage on Windows for any of our existing Miden toolchain components, let alone `midenup` itself.
2. It is likely that we would point people to using WSL rather than a native Windows shell environment, since WSL would likely be required for building components from source anyway, and would allow us to largely use the same instructions for all platforms.
3. None of the maintainers are using WIndows, so we don't have anyone that can actually do the work to support Windows today.

For that reason, I think we may close this PR, but I'd like to get @bobbinth's perspective on Windows support in general.

omerbek:
Thanks for the context. I agree this should not imply native Windows support if the project does not want to support that path today. I'll close this PR and the linked issue rather than keep the change open. If Windows/WSL guidance becomes part of the supported installation story later, it can be handled with docs that match that policy.

PR #234, opened 2026-08-12
fix: add project manifest to midenc commands

### LINKED ISSUE #210: `miden build` fails with "expected input file", but the getting-started tutorial recommends it
The [Your first smart contract](https://docs.miden.xyz/builder/get-started/your-first-smart-contract/create/#building-your-contracts) tutorial instructs users to compile their miden project with `miden build`, but this command currently fails with "expected input file".

Compiling with `cargo miden build` works just fine. I discussed this with @TomasArrachea, and it appears to be an issue with `midenup` rather than the project itself.

It would be helpful to either fix the `miden build` command in `midenup`, or update the tutorial to recommend `cargo miden build` in the meantime.

### How to reproduce it

Create a new project and run `miden build` in the project directory

```bash
miden new my-project
cd my-project/contracts/counter-account
miden build
```

`miden build` fails with an "expected input file" error:

```
info: current toolchain is stable and is installed
error: expected input file

Usage: midenc [OPTIONS] [FILE]

For more information, try '--help'.
Error: failed to execute 'miden build'

Caused by:
    'miden build' failed with status 2
```

Passing an input file explicitly, as the error message suggests, also fails:

```
$ miden build src/lib.rs
info: current toolchain is stable and is installed
Error:   x `rustc` returned an error when compiling the input, see stderr output for
  | more details
Error: failed to execute 'miden build src/lib.rs'
Caused by:
    'miden build src/lib.rs' failed with status 1
```

### Environment

`miden --version` output:

```
The Miden toolchain porcelain:

Environment:
- cargo version: cargo 1.97.0-nightly (eb9b60f1f 2026-04-24).

Midenup:
- midenup + miden version: 0.3.2.
- active toolchain version: 0.15.0.
- midenup revision: .
- midenup was compiled with cargo 1.93.1 (083ac5135 2025-12-15).
```

### LINKED ISSUE #233: bug: `miden build` invokes `midenc`, while `cargo miden build` cannot find managed `cargo-miden`
### Packages versions

midenup: 1.0.0-alpha.1 and 1.0.0-beta.1
Miden toolchain: 0.15.0
cargo-miden: 0.9.0
miden compiler (`midenc`): 0.8.1
Cargo: 1.97.1
Platform: macOS ARM64

### Bug description

After installing Miden toolchain `0.15.0` with midenup, neither available command builds a project:

- `miden build` invokes `midenc` directly and fails because no input file is provided.
- `cargo miden build` fails because Cargo cannot find the midenup-managed `cargo-miden` extension.

midenup installs `cargo-miden` under the active toolchain and creates this shim:

`$MIDENUP_HOME/toolchains/0.15.0/opt/cargo-miden`

However, that directory is added to `PATH` only for commands spawned through the `miden` wrapper. It is not present in the user's shell `PATH`, so a direct `cargo miden build` cannot discover the extension.

Manually running the following makes the build succeed:

`PATH="$MIDENUP_HOME/opt:$PATH" cargo miden build`

This appears related to commit `b57b51c`, which changed the `miden build` alias from `cargo miden build` to `midenc`. Before that change, the wrapper supplied the toolchain PATH while invoking Cargo Miden.

Expected behavior: a project generated with `miden new` should build immediately after installing its toolchain. Either `miden build` should dispatch to the managed `cargo-miden`, or another supported mechanism should make `cargo miden build` discover the correct toolchain-managed extension.

This was reproduced with the currently installed `1.0.0-alpha.1` and with a clean isolated installation of `1.0.0-beta.1`.

Related:

- https://github.com/0xMiden/midenup/issues/159
- https://github.com/0xMiden/midenup/pull/161
- https://github.com/0xMiden/midenup/commit/b57b51cdbc3ba4b50f081a4d27fd43c51362f70c
- https://github.com/0xMiden/docs/issues/343
- https://github.com/0xMiden/docs/pull/352

### How can this be reproduced?

1. Install and initialize midenup:

   `cargo install midenup --version 1.0.0-beta.1`

   `midenup init`

2. Install and select toolchain `0.15.0`:

   `midenup install 0.15.0`

   `midenup override 0.15.0`

3. Generate a project:

   `miden new my-test-project`

4. Enter the generated counter contract:

   `cd my-test-project/contracts/counter-account`

5. Run:

   `cargo miden build`

   Cargo reports that the `miden` subcommand does not exist.

6. Run:

   `miden build`

   This invokes `midenc`, which fails because it was not given an input file.

7. Confirm that Cargo Miden is installed and works when the managed toolchain directory is added manually:

   `PATH="$MIDENUP_HOME/opt:$PATH" cargo miden build`

   The build succeeds and creates:

   `target/miden/dev/counter-account.masp`

### Relevant log output

```shell
$ cargo miden build
error: no such command: `miden`

help: a command with a similar name exists: `miri`

help: view all installed commands with `cargo --list`
help: find a package to install `miden` with `cargo search cargo-miden`

$ miden build
info: current toolchain is stable and is installed
error: expected input file

Usage: midenc [OPTIONS] [FILE]

For more information, try '--help'.
Error: failed to execute 'miden build'

Caused by:
    'miden build' failed with status 2
```

TomasArrachea:
Closes https://github.com/0xMiden/midenup/issues/233.
Closes https://github.com/0xMiden/tutorials/issues/210.

This PR changes `miden build` aliases on 0.15.0 and 0.16.0 to pass `miden-project.toml` as a parameter to midenc. This means that users will not be able to run `miden build` to compile standalone files, it will only be intented to compile a whole project.

### DIFF
--- a/manifest/channel-manifest.json
+++ b/manifest/channel-manifest.json
@@ -1,6 +1,6 @@
 {
   "manifest_version": "3.0.0",
-  "date": 1786461825,
+  "date": 1787603290,
   "networks": {
     "devnet": "0.16.0",
     "mainnet": "0.15.0",
@@ -1547,7 +1547,8 @@
         {
           "aliases": {
             "build": [
-              "%installed-executable"
+              "%installed-executable",
+              "miden-project.toml"
             ]
           },
           "artifacts": {
@@ -1882,7 +1883,7 @@
           ],
           "version": {
             "kind": "registry",
-            "version": "0.29.0"
+            "version": "0.29.2"
           }
         },
         {
@@ -1943,8 +1944,65 @@
         },
         {
           "aliases": {
-            "build": [
+            "format": [
               "%installed-executable"
+            ]
+          },
+          "artifacts": {
+            "miden-format": {
+              "targets": {
+                "aarch64-apple-darwin": {},
+                "x86_64-unknown-linux-gnu": {}
+              },
+              "uri": "https://github.com/0xMiden/miden-vm/releases/download/v%version/miden-format-%target"
+            }
+          },
+          "hide": false,
+          "installation_method": {
+            "kind": "prebuilt"
+          },
+          "installed-executable": "miden-format",
+          "kind": "executable",
+          "name": "format",
+          "profiles": [],
+          "version": {
+            "kind": "registry",
+            "version": "0.29.2"
+          }
+        },
+        {
+          "aliases": {
+            "registry": [
+              "%installed-executable"
+            ]
+          },
+          "artifacts": {
+            "miden-registry": {
+              "targets": {
+                "aarch64-apple-darwin": {},
+                "x86_64-unknown-linux-gnu": {}
+              },
+              "uri": "https://github.com/0xMiden/miden-vm/releases/download/v%version/miden-registry-%target"
+            }
+          },
+          "hide": true,
+          "installation_method": {
+            "kind": "prebuilt"
+          },
+          "installed-executable": "miden-registry",
+          "kind": "executable",
+          "name": "local-registry",
+          "profiles": [],
+          "version": {
+            "kind": "registry",
+            "version": "0.29.2"
+          }
+        },
+        {
+          "aliases": {
+            "build": [
+              "%installed-executable",
+              "miden-project.toml"
             ]
           },
           "artifacts": {
@@ -2183,7 +2241,7 @@
           ],
           "version": {
             "kind": "registry",
-            "version": "0.29.0"
+            "version": "0.29.2"
           }
         }
       ]


manifest/channel-manifest.json
```diff
@@ -1547,7 +1547,10 @@
         {
           "aliases": {
             "build": [
-              "%installed-executable"
+              "%installed-executable",
+              "-L",
+              "%lib",
+              "%project"
```
bitwalker: This could just be `miden-project.toml`

manifest/channel-manifest.json
```diff
@@ -1944,7 +1947,10 @@
         {
           "aliases": {
             "build": [
-              "%installed-executable"
+              "%installed-executable",
```
bitwalker, line 2004: The compiler that ships for 0.16.0 doesn't have the issue that the previous toolchain did (i.e. `miden-project.toml` is inferred if no input file is given)

src/exec.rs
```diff
@@ -109,10 +110,18 @@ pub enum Expr {
     VarPath(Option<String>),
     /// Resolve the command to a file in the toolchain etc directory (`<toolchain>/etc/<file>`).
     EtcPath(String),
+    /// Resolve the command to the project manifest (`miden-project.toml`) in the working directory.
```
bitwalker: I don't think we need this - none of our tooling requires an explicit manifest path today (that was a bug in the previous compiler), and if there are any that require an explicit path, we can simply use the literal `miden-project.toml` string in the manifest as needed

src/exec.rs
```diff
@@ -280,6 +316,25 @@ impl Resolver {
                     None => self.var.clone().into_os_string(),
                 })
             },
+            // Unlike `%lib` and `%etc`, this names a file the *user* owns, so a missing one is a
```
bitwalker: This is not the responsibility of `midenup`

bitwalker (changes requested):
I think this can be reduced to simply using the literal string `"miden-project.toml"` in the channel manifest with the `0.15.0` version of `midenc` - the rest of it is unnecessary IMO.

manifest/channel-manifest.json
TomasArrachea: I think this is actually still not working on the latest compiler version - I created an issue there https://github.com/0xMiden/compiler/issues/1330. So I kept the `miden-project.toml` also for 0.16.0.

bitwalker (approved):

PR #237, opened 2026-08-25
feat: print progress

### LINKED ISSUE #224: RFC: progress reporting for expensive operations
Currently, the work done by `midenup` to install/uninstall/upgrade toolchains and their components is largely invisible - this is because much of the backbone of those operations is completely rewritten, and some of our previous options (e.g. just forwarding `cargo` output) are no longer sufficient for solving the problem.

Installs in general now are very fast - but they still involve downloads of assets, and in some edge cases may require installing/building things with `cargo`. It would be ideal if we could provide nice progress reporting in cases where it is requested, that makes it easy to understand what is happening behind the scenes (and in perhaps some cases, being able to request the raw output for troubleshooting).

IMO, this calls for four modes:

* `--quiet` never prints any progress at all
* default output simply prints an informational message that mentions what `midenup` is doing, when it is about to do something expensive, with an optional percentage, or N/M unit-based progress (e.g. `installing channel 'foo'...10%` or `upgrading components of channel 'foo'...5/10`)
* `--progress` prints out every download/install task that is being performed and its progress (i.e. per-component, not one progress for the whole toolchain)
* `--verbose` prints out the raw output of the tasks being performed, i.e. for a download, we emit the progress report for the download, for a `cargo`-based operation we pipe the output from `cargo` directly. If `-vv` is given, we further print debug output that describes every action that is being taken, including low-level file system operations.

This provides useful options for both common use cases, and uncommon troubleshooting use cases.

TomasArrachea:
Closes https://github.com/0xMiden/midenup/issues/224.

This PR adds prints that show the progress during midenup operations. There are 4 levels of verbosity:
- `-q`: warnings and errors only. No progress, no announcements.
- default: on installations shows one line per component as it is installed, plus a live transfer display on a terminal.
- `-v`: The above, and the output of spawned programs (e.g. cargo) is no longer suppressed.
- `-vv`: The above, and every action `midenup` takes is traced.

<details>
<summary>Example default installation</summary>

midenup-install-0 16
</details>

<details>
<summary>Example verbose (`-v`) installation</summary>
midenup-install-0 16-build-verbose

</details>

<details>
<summary>Example debug (`-vv`) installation</summary>
midenup-install-0 15-debug

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,9 +60,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.103"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "ascii"
@@ -103,9 +103,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.65"
+version = "1.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e228eec9be7c17ccb640b59b36a5cd805ea2a564a4c5e162c2f659fea30d3b96"
+checksum = "0ad534f4357a5264cce5019c989cf66a4f0dc4e0d1b1d15f8aacec0ff7360273"
 dependencies = [
  "find-msvc-tools",
  "shlex",
@@ -134,9 +134,9 @@ checksum = "6e4de3bc4ea267985becf712dc6d9eed8b04c953b3fcfb339ebc87acd9804901"
 
 [[package]]
 name = "clap"
-version = "4.6.1"
+version = "4.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
+checksum = "473c7e07f409a8d772161724aa8db6a765a2532a70f9667eeb7b49d3d02fbdca"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -144,9 +144,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.6.0"
+version = "4.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "714a53001bf66416adb0e2ef5ac857140e7dc3a0c48fb28b2f10762fc4b5069f"
+checksum = "7b48fea5a88e9ae728a2dcbedbfc0e730f7d60da42e1cb049a83c9fb8b789889"
 dependencies = [
  "anstream",
  "anstyle",
@@ -156,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "4.6.1"
+version = "4.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2ce8604710f6733aa641a2b3731eaa1e8b3d9973d5e3565da11800813f997a9"
+checksum = "d012d2b9d65aca7f18f4d9878a045bc17899bba951561ba5ec3c2ba1eed9a061"
 dependencies = [
  "heck",
  "proc-macro2",
@@ -237,9 +237,9 @@ dependencies = [
 
 [[package]]
 name = "curl-sys"
-version = "0.4.89+curl-8.20.0"
+version = "0.4.90+curl-8.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d680779285438f2d0927485973ab45b212ea990bddb80de8a55a1e3c1d9ba22"
+checksum = "97799a0d220bfb3361e0fe4936966ff8c4b24d65c3f06dfc70d7b680b44e7897"
 dependencies = [
  "cc",
  "libc",
@@ -321,9 +321,9 @@ dependencies = [
 
 [[package]]
 name = "find-msvc-tools"
-version = "0.1.9"
+version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
+checksum = "d45db016d36b838f563236e9193d0ee6ce38f3f68b6c94e914b4929c96bbb890"
 
 [[package]]
 name = "fixedbitset"
@@ -403,9 +403,9 @@ checksum = "df3b46402a9d5adb4c86a0cf463f42e19994e3ee891101b1841f30a545cb49a9"
 
 [[package]]
 name = "hybrid-array"
-version = "0.4.12"
+version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9155a582abd142abc056962c29e3ce5ff2ad5469f4246b537ed42c5deba857da"
+checksum = "707114b52a152fa7bdb290cd7cd5912d9467273b6d74e21b8d81aca1f8533f6b"
 dependencies = [
  "typenum",
 ]
@@ -456,15 +456,15 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.186"
+version = "0.2.189"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68ab91017fe16c622486840e4c83c9a37afeff978bd239b5293d61ece587de66"
+checksum = "3eaf3ede3fee6db1a4c2ee091bf8a8b4dccdc6d17f656fb07896ee72867612f2"
 
 [[package]]
 name = "libredox"
-version = "0.1.17"
+version = "0.1.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f02ab6bace2054fb888a3c16f990117b579d14a3088e472d63c6011fa185c9d3"
+checksum = "28d0a00925a9f930d679b6789b721e3a7f9ed110f41b86d2497caa780c3a070a"
 dependencies = [
  "libc",
 ]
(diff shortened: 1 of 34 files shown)


docs/src/design/spec.md
```diff
@@ -1050,6 +1049,42 @@ Each variant carries the file path, the offending identifier, and a remediation
...
+### 14.4 Output and verbosity
+
+**Streams.** Stdout carries a command's results; stderr carries its progress, warnings, and traces.
+
+**Levels.** Four, ordered:
+
+| Level | Flag | Emits |
+|---|---|---|
+| quiet | `-q`, `--quiet` | warnings and errors only |
+| normal | *(default)* | one line per unit of work as it happens |
```
bitwalker: I would make this a bit more granular, with the following flags and interactions, expressed below in derive-style `clap` example form).

These flags ensure that common CLI flags are represented, and provides some flexibility in output:

* `--quiet` suppresses output almost entirely, and the value of `--verbosity` is inferred from it, so that one does not have to specify both. They are also marked as conflicting so as to avoid non-sensical combinations.
* `--plain` sets defaults to avoid terminal decorations that may not render properly to the pseudoterminal
* Set "do-what-I-mean" default missing values for flags, e.g. `--progress` is equivalent to `--progress=pretty`, since the user is clearly requesting progress be displayed. Since we choose defaults conditioned on other flags, this still interacts well with flags like `--plain`, which fill force the value to `--progress=plain`
* The defaults are equivalent to `--progress --color --verbosity=info`, which optimizes for common interactive use.

Examples:

* `--verbose --plain` emits debug output suitable for CI or other non-interactive shell environments
* `--no-progress` suppresses noisy progress output without suppressing all useful informational output

```rust
struct Args {
    /// Emit simple textual output - no color or decorations (e.g. progress bars)
    #[arg(long)]
    plain: bool,
    #[arg(
        long, 
        value_enum,
        value_name = "STYLE",
        conflicts_with("no_progress"),
        num_args(0..=1),
        default_value_t = ProgressStyle::Pretty,
        default_missing_value = "pretty",
        default_value_ifs([
            ("quiet", ArgPredicate::IsPresent, Some("none")),
            ("no_progress", ArgPredicate::IsPresent, Some("none")),
            ("plain", ArgPredicate::IsPresent, Some("plain")),
            ("verbosity", ArgPredicate::IsPresent, Some("pretty")),
        ])
    ]
    progress: ProgressStyle,
    #[arg(
        long,
        default_value_t = false,
        default_value_ifs([
            ("quiet", ArgPredicate::IsPresent, Some("true")),
            ("progress", "none".into(), Some("true")),
            ("progress", ArgPredicate::IsPresent, Some("false")),
            ("verbosity", ArgPredicate::IsPresent, Some("false")),
        ])
    )]
    no_progress: bool,
    #[arg(
        long = "verbose",
        short = 'v',
        value_enum,
        value_name = "LEVEL",
        default_value_t = Verbosity::Info,
        default_missing_value = "info",
        default_value_ifs([
            ("quiet", ArgPredicate::IsPresent, Some("warn")),
        ]),
        num_args(0..=1),
        help_heading = "Diagnostics"
    )]
    verbosity: Verbosity,
    #[arg(long, short = 'q', conflicts_with("verbosity"))]
    quiet: bool,
    #[arg(
        long,
        value_enum,
        default_value_t = ColorChoice::Auto,
        default_missing_value = "auto",
        num_args(0..=1),
        default_value_ifs([
            ("plain", ArgPredicate::IsPresent, Some("false"))
        ])
    )]
    color: ColorChoice,
}


pub enum ColorChoice {
    Auto,
    True,
    False,
}

pub enum ProgressStyle {
    Pretty,
    Plain,
    None,
}

pub enum Verbosity {
    Error,
    Warn,
    Info,
    Debug,
}
```

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+    io::{IsTerminal, Write},
+    sync::atomic::{AtomicU8, Ordering},
+    time::{Duration, Instant},
+};
+
+use colored::Colorize;
+
+/// How much `midenup` says about what it is doing.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Verbosity {
```
bitwalker: This enum conflates too many details, I would structure it like I described in my other comment, where the verbosity level corresponds to the level of information displayed (so `--quiet` corresponds to `Warn`, the default is `Info`, and verbose output is `Debug`).

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Warnings and errors only. No progress, no announcements.
+    Quiet,
+    /// One line per component as it is acquired, plus a live transfer display on a terminal.
+    #[default]
+    Normal,
+    /// The above, and the output of spawned programs (e.g. cargo) is no longer suppressed.
+    Verbose,
+    /// The above, and every action `midenup` takes is traced.
+    Debug,
```
bitwalker: If we need to distinguish between `Verbose` and `Debug`, then I would use `Debug` for  your`Verbose`, and `Trace` for your `Debug` - that corresponds to common levels in various logging frameworks, and better conveys the level of detail.

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+            _ => Self::Debug,
+        }
+    }
+}
+
+/// The level in effect, as a [Verbosity] discriminant.
+///
+/// Process-global because the things that report are not all reachable from a [crate::config]:
+/// [crate::lock] is handed a path, and the executors are handed a plan.
+static LEVEL: AtomicU8 = AtomicU8::new(Verbosity::Normal as u8);
```
bitwalker: `Verbosity` needs to be explicitly defined as `#[repr(u8)]` if you are going to do this.

bitwalker: On second thought, why not just thread this through using `Config`? It isn't really clear to me why it needs to be process global - all the places that would need to emit output to the user are going to be near a `Config` anyway - it might be a bit more invasive, but it would avoid the need for global state like this.

I'm not entirely opposed to the idea, but it does make testing significantly uglier/more difficult, because every test shares the same global state.

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+/// Whether a live, redrawn line for a long-running child process is appropriate.
+///
+/// Only at exactly [Verbosity::Normal]: above it the child's own output is shown, and a line
+/// redrawn underneath would interleave with it.
+pub fn activity_is_live() -> bool {
+    verbosity() == Verbosity::Normal && std::io::stderr().is_terminal()
+}
+
+/// Emits a labelled line at [Verbosity::Normal] and above. Prefer the [crate::info] macro.
+pub fn emit_info(args: fmt::Arguments) {
```
bitwalker: By taking `fmt::Arguments`, most of the work in terms of formatting happens even if the message never gets written - it's not _all_ of the work, but it's still non-trivial.

I would suggest making all of these macros, and having the macro do the gating at the call site - that will avoid doing much of the work to emit, say, trace messages, even though those messages will never be rendered in the usual case.

bitwalker (changes requested):
I'm still reviewing, but had some early feedback I wanted to share. I'll follow up with the rest of my review ASAP

src/commands/update.rs
```diff
@@ -27,6 +27,9 @@ pub fn update(
     state: &mut LocalState,
     options: &UpdateOptions,
 ) -> anyhow::Result<()> {
+    // Sync up front, so the header precedes any per-channel work; later calls hit the cache.
+    config.upstream_manifest()?;
```
bitwalker: This sync now runs before the empty-installation and pinned-version existence checks. As a result, `midenup update` on a fresh or offline machine fails with a manifest-fetch error instead of reporting that there is nothing to update, and a missing pinned version can have its local diagnostic masked by the network failure. Please move the fetch after the local no-work and validation branches; the network-selector path can continue fetching inside `update_network`.

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+    // skip if this is not a terminal, we don't need transfer data for file/CI logs.
+    verbosity() >= Verbosity::Normal && std::io::stderr().is_terminal()
+}
+
+/// Whether a live, redrawn line for a long-running child process is appropriate.
+///
+/// Only at exactly [Verbosity::Normal]: above it the child's own output is shown, and a line
+/// redrawn underneath would interleave with it.
+pub fn activity_is_live() -> bool {
+    verbosity() == Verbosity::Normal && std::io::stderr().is_terminal()
```
bitwalker: At normal verbosity this enables a no-newline activity display on the same inherited stderr used by the Cargo and extraction children. Cargo's `--quiet` flag still allows compiler errors through, so redraws can append to, overwrite, or erase those diagnostics. Please capture and multiplex child stderr so it can be replayed after the activity line is cleared, or disable live activity whenever child stderr is inherited.

src/report.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+/// Emits a labelled line at [Verbosity::Normal] and above. Prefer the [crate::info] macro.
+pub fn emit_info(args: fmt::Arguments) {
+    if verbosity() >= Verbosity::Normal {
+        write_line(format_args!("{}: {args}", "info".bold()));
```
bitwalker: These labels are written to stderr, but `colored`'s automatic policy checks whether stdout is a terminal. This emits ANSI escapes into redirected stderr when stdout remains attached to a terminal, and disables color on interactive stderr when stdout is piped. Please make automatic color selection depend on the stream being styled, while keeping explicit plain/color overrides orthogonal.

docs/src/design/spec.md
```diff
@@ -1050,6 +1049,42 @@ Each variant carries the file path, the offending identifier, and a remediation
...
+| verbose | `-v` | the above, and the output of spawned programs is no longer suppressed |
+| debug | `-vv` | the above, and the individual actions taken: fetches, spawned commands, seeded files, symlink commits, link updates, record writes, deletions |
```
bitwalker: This promises that the individual actions are traced, but substantial mutation paths are currently untraced. For example, a fresh `midenup -vv init` creates the home layout and Cargo-bin symlink without emitting any `debug:` line; default-link replacement and `opt` link repair have similar gaps. Please either add traces and command-specific coverage for the promised mutation classes, or narrow this contract to the actions actually reported.

docs/src/design/spec.md
```diff
@@ -1050,6 +1049,42 @@ Each variant carries the file path, the offending identifier, and a remediation
 
 `DivergentState` and `NeedsReinstall` both name the exact recovery command.
 
+### 14.4 Output and verbosity
+
+**Streams.** Stdout carries a command's results; stderr carries its progress, warnings, and traces.
```
bitwalker: The interactive path-update flow still uses `println!` for its prompt and update/cancel/skip acknowledgements, so `--path-update=interactive` writes non-result text to stdout despite this contract. Please route the prompt and status text to stderr and flush the prompt before reading stdin, leaving stdout reserved for command results.

src/install/stage.rs
```diff
@@ -193,6 +204,106 @@ pub fn execute(
...
+                "downloading" => ("download", "downloads"),
+                "copying" => ("copy", "copies"),
+                "building" => ("source build", "source builds"),
+                _ => ("extraction", "extractions"),
+            };
+            Some(format!("{count} {}", if count == 1 { singular } else { plural }))
+        })
+        .collect();
+
+    crate::info!("{} steps: {}", planned.len(), described.join(", "));
```
bitwalker: This always pluralizes `steps`, so a one-unit installation reports output such as `1 steps: 1 copy`. Please select `step` or `steps` from `planned.len()` just as the per-kind counts are singularized above.

Cargo.lock
```diff
@@ -60,9 +60,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.103"
+version = "1.0.104"
```
bitwalker: There is no corresponding dependency-declaration change in this PR, but the lockfile advances 27 packages, including a major `syn` update. Please revert this unrelated lockfile churn or split it into a separate, justified dependency update so the progress-reporting change does not absorb unrelated compatibility and supply-chain risk.

bitwalker (changes requested):
Ok, finished up review, and Codex caught a few in addition to some of my own findings. See comments for details. Once these are addressed we can get this merged. Nice work!

bitwalker (approved):

PR #239, opened 2026-08-28
fix: partial install detection

TomasArrachea:
Display fixes for `midenup show list` command:
- `(partially installed)` no longer gets shown for every default install. The marker was derived by comparing with the channel's complete component set, which minimal installs never match. It is now derived by resolving the recorded install intent against upstream. It would show if a channel added a component after the user installed it.
- Network drift is reported per network, under the list. Before, the message was appended to the installed channel entry which made it confusing.

### DIFF
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -11,11 +11,11 @@ pub fn list(config: &Config, state: &LocalState) -> anyhow::Result<()> {
             let channel_name = &channel.name;
 
             // Partial status is *derived*, never recorded (spec section 8.6): an installation is
-            // partial exactly when it holds fewer components than the channel offers. A stored
-            // flag would be a second answer to a question the component set already answers, and
-            // the two would drift.
+            // partial exactly when upstream resolves its recorded intent to components it does
+            // not hold. A stored flag would be a second answer to a question the component set
+            // already answers, and the two would drift.
             let installed_indicator = match state.get(&channel.name) {
-                Some(installation) if installation.as_channel().is_partially_installed(channel) => {
+                Some(installation) if installation.is_partially_installed(channel) => {
                     format!(" {}", "(partially installed)".yellow())
                 },
                 Some(_) => format!(" {}", "(installed)".green()),
--- a/docs/src/design/spec.md
+++ b/docs/src/design/spec.md
@@ -558,7 +558,7 @@ the superset is a warning, and the component in the active view wins.
 
 ### 8.6 Partial status is derived
 
-Local state does not record a "partial" flag. When the upstream manifest is available, an installation is displayed as partial if its installed component set is a proper subset of the channel's complete set. When upstream is unavailable, partial status is not displayed.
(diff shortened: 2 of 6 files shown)


src/commands/update.rs
```diff
@@ -409,6 +409,56 @@ fn work_for(
...
+/// Whether two definitions of one component differ as manifest content.
+///
+/// The pins install records on an authority -- a branch's commit, a path's modification time --
+/// exist only locally and move on their own, so they are normalized away: drift behind them is
+/// reconciled by the update itself, which is what pins for ([`classify`]); a listing must not.
+fn definition_changed(installed: &Component, upstream: &Component) -> bool {
+    let normalized = |component: &Component| {
+        let mut component = component.clone();
+        match &mut component.version {
+            Authority::Path { last_modification, .. } => *last_modification = None,
```
bitwalker: **[P2] Normalize relative path authorities before comparing definitions**

Installation rewrites a relative `Authority::Path.path` using `config.working_directory.join(path)` in `refresh_path_modification_times`, but this comparison only removes `last_modification`. Consequently, an unchanged manifest with `"version": {"kind": "path", "path": "src"}` shows `(update available)` in both `list` and `show list` immediately after a successful install: the stored absolute path differs from the raw relative path. Running `update 0.15.0 --path-update=all` succeeds but never clears the marker, because metadata updates preserve the installed authority. This was reproduced against this PR's head. Please account for the install-time path transformation when comparing definitions, while preserving the requirement that listing does not access component sources.

src/commands/list.rs
```diff
@@ -11,15 +11,24 @@ pub fn list(config: &Config, state: &LocalState) -> anyhow::Result<()> {
...
+                Some(installation) if super::update::needs_update(installation, channel) => {
+                    format!(" {}", "(update available)".yellow())
                 },
                 Some(_) => format!(" {}", "(installed)".green()),
+                // A channel that supersedes an installed one (spec section 11.4): updating the
+                // predecessor migrates to it.
+                None if channel
+                    .migrates_from
+                    .as_ref()
+                    .is_some_and(|old| state.get(old).is_some()) =>
```
bitwalker: **[P2] Honor same-version precedence when displaying migration updates**

Checking only whether the predecessor is installed does not match the updater's counterpart-selection rule. If upstream contains both installed `0.15.0` and an uninstalled `0.16.0` declaring `migrates_from: "0.15.0"`, this branch prints `0.16.0 (update available)`. However, `upstream_counterpart_raw` prefers the still-published `0.15.0`, so `update 0.15.0` reports `Toolchain 0.15.0 is up to date` and performs no migration; `show list` also reports no update. This was reproduced against this PR's head, and the false marker remains after updating. Please use the same counterpart-selection rule as the updater when deciding whether a successor represents an available update.

bitwalker (changes requested):
Looks good, but Codex found two issues that check out that we should fix - see comments for details

bitwalker (approved):

PR #241, opened 2026-08-29
manifest(0.15.0): fix `deploy` alias

### LINKED ISSUE #240: `miden deploy` still broken on 0.15.0: `-s` is passed to `miden client`
This was item (3) in 0xMiden/compiler#1349; bitwalker suggested it belongs here.

**Short version: the fix only landed on 0.16.0.** `networks` still points `mainnet` and `testnet` at 0.15.0, so on the default toolchain nothing has changed:

```console
$ miden deploy
info: current toolchain is mainnet and is installed
error: unexpected argument '-s' found

Usage: miden client [OPTIONS] <COMMAND>

Error: failed to execute 'miden deploy'

Caused by:
    'miden deploy' failed with status 2
```

Looking at the published manifest, 0.9.0 through 0.15.0 carry no subcommand at all:

```json
["%installed-executable", "-s", "public", "--account-type", "regular-account-immutable-code"]
```

which is what puts `-s` on `miden client` itself. Even with a subcommand in front of it the line wouldn't work on 0.15.0 — the type flag there is `-t`, and `--account-type` takes only `private` or `public`.

0.16.0 has the working form (3987812):

```json
["%installed-executable", "new-account", "--account-type", "public", "--deploy"]
```

and it looks safe to backport. On 0.15.0, `miden client new-account -t public --deploy -p <missing file>` exits 1 on the missing package instead of 2 on argument parsing, so the flags are accepted there. I only checked 0.15.0 though, so no idea how far back the same line can go.

midenup reports `0.15.0 (mainnet, testnet)`, active toolchain `mainnet`, Linux x86_64.

Mustdzyl:
Fixes #240.

`deploy` on 0.15.0 hands `-s` straight to `miden client`, so the alias never parses:

```console
$ miden deploy
error: unexpected argument '-s' found
```

This copies the 0.16.0 line (3987812) onto 0.15.0, which is what both `mainnet` and `testnet` point at today. I left the older channels alone, since I only checked `new-account` against 0.15.0.

Verified on an installed 0.15.0 toolchain: with the alias patched in local state, `miden deploy --help` reaches `new-account`'s help instead of the parse error.

Worth noting for anyone already installed — aliases are snapshotted into `state.json` at install time, so the manifest change on its own won't heal an existing installation. `midenup update` has to re-resolve it, which it does, since an alias-only change classifies as `RuntimeMetadataOnly`.


bitwalker (changes requested):
Looks good, but you'll need to run the following locally and squash the change in to your commit:

```
make update-manifest
bin/update-manifest --manifest-path manifest/channel-manifest.json touch
```

This updates the timestamp that will cause existing installs to pick up the change.

Mustdzyl:
Done, squashed into the same commit. `check` passes on the result.

That also answers the question I left on #240 — I could not see how an existing install would ever pick the alias up, and the timestamp is the piece I was missing.

bitwalker (approved):

